PR #200, opened 2013-07-03
remove ZF2 submodule

ghost:
Zend Framework is being installed both by Composer and git submodule.
It makes confusion with two ZF's (/vendor/ZF2 and /vendor/zendframework).

As Composer is recommended way to deploy skeleton, let's remove the submodule.

Even if developer does not run Composer to install skeleton,
he may want to install another libraries with it.
Those libaries (e.g. Doctrine) are dependant on zendframework/zendframework
package, so it will result in two ZF's anyways.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "vendor/ZF2"]
-	path = vendor/ZF2
-	url = https://github.com/zendframework/zf2.git
--- a/README.md
+++ b/README.md
@@ -44,7 +44,8 @@ Using Git submodules
 --------------------
 Alternatively, you can install using native git submodules:
 
-    git clone git://github.com/zendframework/ZendSkeletonApplication.git --recursive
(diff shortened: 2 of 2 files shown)


chatzipan:
Hi guys,

I have an issue which might be related with this one, so I don't want to open a new one. I've been using zf2 for about a year, and in the agency where Im working, there are many zf2 based projects.

Now I just noticed today that in ALL off these projects the init_autoloader.php script is loading the framework from my globally installed library which is located under /usr/local/zend/share/ZendFramework2/library and not from vendor/zendframework where Composer had installed the library when i setup the project. 

Any idea why this is happening? this shouldn't be happening right?

Maks3w:
With ZF 2.5 component splitting this may be reopened.

chatzipan:
I mean the script keeps looking for if(is_dir('vendor/ZF2/library')){

but I don't have such a directory under vendor only  vendor/zendframework . But the app can run because it was using the framework installed globally on my laptop..

PR #238, opened 2014-01-29
Fixed locale files (accept-language uses dashes instead of underscores)

wimg:
Discovered this issue during Evan's training. This should enable usage of locale from the browser's accept-language header


akrabat:
According to [RFC 4646](http://www.ietf.org/rfc/rfc4646.txt), the separator is a hyphen. 

It also says:

>The tags and their subtags, including private use and extensions, are
   to be treated as case insensitive: there exist conventions for the
   capitalization of some of the subtags, but these MUST NOT be taken to
   carry meaning.

akrabat:
Needs rebasing and bringing up to date with new translations that have been added.
Please reopen if you do this work and I'll merge.

PR #240, opened 2014-02-03
Add composer.lock to git

jmleroux:
composer.lock should be in VCS.

`php composer.phar install` will be faster

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,5 +6,4 @@ nbproject
 .idea
 .project
 .settings
-composer.lock
-vendor/bin
+vendor/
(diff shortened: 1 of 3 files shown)


jeremiahsmall:
> Please read previous discussions about that.

@Ocramius is prior discussion stale now? @weierophinney and @ezimuel decided to ship `composer.lock` and not ship `compooser.phar` in the Apgility skeleton. They also suggested same should apply to this skeleton. See https://github.com/zfcampus/zf-apigility-skeleton/pull/92#issuecomment-110063604

akrabat:
Implemented as part of #286

PR #245, opened 2014-03-05
en_US.mo file must not contain empty strings

bartko-s:
Translator object config
```
$translator->setLocale('en_US')
           ->setFallbackLocale('cs_CZ');
```

View script
```
echo $this->translate('A 404 error occurred');
```
Expected output is "A 404 error occurred" but actual output is "Vyskytla se chyba 404" which is incorrect.

This pull request fix this issue.


akrabat:
Patch is no longer available.

PR #255, opened 2014-03-21
Hotfix/end loops exception

netroby:
The exception dump loops should be end at some where, or your memory should be not enough to handle the request.


netroby:
My tears fly on my face after @akrabat merged this PR. I was waiting for so long time.

PR #269, opened 2014-09-03
Updates to latest (supported) versions of Bootstrap, jQuery & html5shiv

rahuldroy:


akrabat:
We've been so slow on this that Bootstrap is now 3.3.4!

I've cherry picked the jQuery and html5shiv updates, but not the Bootstrap. If you're minded to provide a PR to update Bootstrap to 3.3.4, I'll take it. 

Note that each PR should be on its own branch rather than directly from your master branch.

PR #272, opened 2014-10-21
Update README.md

bjohnmer:
You have to add the Require all granted line at the apache configuration, because this have to be enabled in order to it could resolve vhost succesfully. I have Ubuntu Server 14.04 with Apache/2.4.7 many guys would appreciate that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,5 +73,11 @@ project and you should be ready to go! It should look something like below:
             AllowOverride All
             Order allow,deny
             Allow from all
+            #
+            # NOTICE: 
+            # You have to write
+            Require all granted
+            # in order to Apache could resolve ZF2 vhosts
+            #
         </Directory>
     </VirtualHost>


README.md
```diff
@@ -73,5 +73,11 @@ project and you should be ready to go! It should look something like below:
             AllowOverride All
             Order allow,deny
             Allow from all
+            #
+            # NOTICE: 
+            # You have to write
+            Require all granted
+            # in order to Apache could resolve ZF2 vhosts
+            #
```
jeremiahsmall: Probably doesn't need all the comments inline.

akrabat:
Require all granted is Apache 2.4+, so the correct code is something like this:

```
<IfModule !mod_authz_core.c>
Order deny,allow
Allow from all
</IfModule>
<IfModule mod_authz_core.c>
Require all granted
</IfModule>
```

akrabat:
Though maybe we can assume that no one is running 2.2 anymore?!

akrabat:
Fixed as part of #286.

PR #279, opened 2015-03-13
update zendframework dependency to ~2.3

samsonasik:
it can be merge when zf 2.4 released.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,6 +9,6 @@
     "homepage": "http://framework.zend.com/",
     "require": {
         "php": ">=5.3.3",
-        "zendframework/zendframework": "2.3.*"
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -9,6 +9,6 @@
     "homepage": "http://framework.zend.com/",
     "require": {
         "php": ">=5.3.3",
-        "zendframework/zendframework": "2.3.*"
+        "zendframework/zendframework": "2.4.*"
```
Ocramius: Should probably be `~2.3`

samsonasik: @Ocramius ok, updated.

Ocramius:
:+1:

PR #281, opened 2015-04-20
update composer.json to ~2.4

samsonasik:
based on http://framework.zend.com/long-term-support

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,6 +9,6 @@
     "homepage": "http://framework.zend.com/",
     "require": {
         "php": ">=5.3.3",
-        "zendframework/zendframework": "~2.3"
+        "zendframework/zendframework": "~2.4.0"
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -9,6 +9,6 @@
     "homepage": "http://framework.zend.com/",
     "require": {
         "php": ">=5.3.3",
-        "zendframework/zendframework": "~2.4.0"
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -9,6 +9,6 @@
     "homepage": "http://framework.zend.com/",
     "require": {
         "php": ">=5.3.3",
-        "zendframework/zendframework": "~2.3"
+        "zendframework/zendframework": "~2.4.0"
```
Ocramius: `~2.4`

samsonasik:
@Ocramius done ;)

samsonasik:
@Ocramius  @weierophinney  merge-able ?

Ocramius:
:+1:

samsonasik:
@Ocramius when using ~2.4, I got 2.5.1, it is that purpose ? or we should use ~2.4.0 instead ?

Ocramius:
@samsonasik `~2.4` means `>=2.4,<3.0`

samsonasik:
ok

DonCarlosGraz:
hi! 

I noticed that there is no 2.4.x package on 
https://packages.zendframework.com/

Is this planned in the near future?

Thanks in advance

Maks3w:
@DonCarlosGraz Please open your issue in https://github.com/zendframework/ZF2Package

PR #283, opened 2015-04-26
Updated public/index.php, exclude cli-server static file requests to self

bacinsky:
Fixed cli-server static file requests for URI with /index.php

### DIFF
--- a/public/index.php
+++ b/public/index.php
@@ -6,9 +6,11 @@
 chdir(dirname(__DIR__));
 
 // Decline static file requests back to the PHP built-in webserver
-if (php_sapi_name() === 'cli-server' && is_file(__DIR__ . parse_url($_SERVER['REQUEST_URI'], PHP_URL_PATH))) {
+$path = __DIR__ . parse_url($_SERVER['REQUEST_URI'], PHP_URL_PATH);
+if (php_sapi_name() === 'cli-server' && __FILE__ !== $path && is_file($path)) {
     return false;
 }
+unset($path);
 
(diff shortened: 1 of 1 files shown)


public/index.php
```diff
@@ -6,9 +6,11 @@
 chdir(dirname(__DIR__));
 
 // Decline static file requests back to the PHP built-in webserver
-if (php_sapi_name() === 'cli-server' && is_file(__DIR__ . parse_url($_SERVER['REQUEST_URI'], PHP_URL_PATH))) {
+$path = __DIR__ . parse_url($_SERVER['REQUEST_URI'], PHP_URL_PATH);
+if (php_sapi_name() === 'cli-server' && __FILE__ !== $path && is_file($path)) {
```
Ocramius: Can `$path` be different from `__FILE__` and still be equivalent (symlink)? `realpath()` it?

samsonasik:
@Ocramius mergeable now?

public/index.php
```diff
@@ -6,9 +6,11 @@
 chdir(dirname(__DIR__));
 
 // Decline static file requests back to the PHP built-in webserver
-if (php_sapi_name() === 'cli-server' && is_file(__DIR__ . parse_url($_SERVER['REQUEST_URI'], PHP_URL_PATH))) {
+$path = realpath(__DIR__ . parse_url($_SERVER['REQUEST_URI'], PHP_URL_PATH));
```
akrabat: Move the realpath() test  inside the test for the cli-server, please.

Issue #62, opened 2012-05-22
Remove composer.phar


Seldaek:
Please do not distribute composer.phar along with the app, because that means people don't run system checks and then we get support requests for stuff that would have been avoided if they ran the installer from http://getcomposer.org/download/ /cc @SpiffyJr

PR #286, opened 2015-06-09
Composer changes

jeremiahsmall:
Add composer.lock and remove composer.phar per this discussion: https://github.com/zfcampus/zf-apigility-skeleton/pull/92#issuecomment-109941631
(Moving the Vagrant addition to its own PR)


README.md
```diff
@@ -13,24 +13,37 @@ Installation
 Using Composer (recommended)
 ----------------------------
 The recommended way to get a working copy of this project is to clone the repository
-and use `composer` to install dependencies using the `create-project` command:
+and use `composer` to install dependencies using the `install` command:
 
-    curl -s https://getcomposer.org/installer | php --
-    php composer.phar create-project -sdev --repository-url="https://packages.zendframework.com" zendframework/skeleton-application path/to/install
+```
+git clone https://github.com/zfcampus/zf-apigility-skeleton.git # optionally, specify the directory in which to clone
```
samsonasik: zf-apigility?

jeremiahsmall:
@samsonasik thanks for spotting the copy/paste issue with the github url in the readme.

RobQuistNL:
:+1: the proper way.

gianarb:
:+1: for composer.lock..
Composer phar into the VCS was added for an old discussion and for help a speedy project startup..

I agree with you but this implementation has already discussed

RobQuistNL:
@gianarb whats so speedy about it? Its outdated, you even have to run self-update before running it. This downloads a new phar, just like the updated description does. If any, this is faster, because the user won't have to download the outdated .phar at all :)

gianarb:
composer.phar is outdated is because this project has the problem of maintenance..
obviously it borns to be always updated

RobQuistNL:
Then why does the README.md tells you to run "self-update"? Composer is very well maintained and updates to a propery release pretty often, so the self-update would download a new version very often. Honestly, I think the composer.phar in the project adds a little confusion for new users. Especially since it will show up as a change after the self-update.

gianarb:
Oh yes I agree with you composer.phar into the VCS creates confusion and is
a bad practice IMO :) I hope that this issue resolve
this misunderstanding..
But this argument is already discussed here
https://github.com/zendframework/ZendSkeletonApplication/issues/62


2015-06-09 10:22 GMT+02:00 Rob Quist <notifications@github.com>:

> Then why does the README.md tells you to run "self-update"? Composer is
> very well maintained and updates to a propery release pretty often, so the
> self-update would download a new version very often. Honestly, I think the
> composer.phar in the project adds a little confusion for new users.
> Especially since it will show up as a change after the self-update.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/zendframework/ZendSkeletonApplication/pull/286#issuecomment-110272778>
> .
>



-- 
Gianluca Arbezzano
www.gianarb.it

akrabat:
I'm +1 for removing composer.phar. It made sense back in 2012, but doesn't any more. 

Ping @weierophinney for his thoughts as he had the last word on #62.

ezimuel:
I discussed with @weierophinney and we both agree, we are going to remove `composer.phar` in the skeleton applications now and for the future. Having a correct README that shows how to install composer is the correct solution.

gianarb:
:+1: Beautiful!

akrabat:
@jeremiahsmall In the README, please put back the `create-project` method as this is the best way to install it. (Note, rebase from master first as I've just merged an improvement to the create-project command.)

Then remove the part about cloning manually and then running composer as there's no point in doing it this way given that we now expect the user have a composer installed.

gianarb:
@ezimuel have you discussed about the vendor this?  
Maybe we can remove even this directory

jeremiahsmall:
I've updated the README to include the `create-project` as Option 1, however, I've also shown how to do the same without curl as Option 2, and I've added steps to copy your newly minted composer.phar into your project afterward. I've also included two additional options. Option 3 is the via curl and tar option that was already in the README, and Option 4 is manual git clone and non-curl composer installation.

For your reading convenience: https://github.com/jeremiahsmall/ZendSkeletonApplication/blob/hotfix/composer-changes/README.md

Frankly, installing composer.phar first outside the project and then using it for `create-project` seems a bit more cumbersome to me, because it then either gets abandoned in some arbitrary directory, and composer still has to be installed inside the project, or it has to be moved into the project (as I've suggestion in options 1 & 2).

akrabat:
I'm curious. Why do you think that composer has to be installed inside the project? It's perfectly happy living in /usr/local/bin.

jeremiahsmall:
@akrabat, it's not that it *has* to be installed in the project, or anywhere in particular for that matter. It's just that in the "preferred" recipe it suggests you run `curl -s https://getcomposer.org/installer | php` which simply downloads the phar anywhere you happen to be. I think it's out of scope for a README like this to try to explain best practices for how/where to install composer.

Based on that, and on the fact that we can't count on curl, I'm actually in favor of only showing the Option 4 recipe.

akrabat:
Merged after rebasing.

jeremiahsmall:
Thx. Small typo in the word "donload" on revised README.md

jeremiahsmall:
Also, still not clear why we put curl in the default instructions when there is another equally simple alternative with readfile.

akrabat:
Fixed Typo. I picked curl as that's the first option in the Composer documentation and was already in the README. One way as an example is enough with a link to the documentation. We don't document using Composer-Setup.exe either, or installation on an IBM i-series, so we don't need to cover all bases.

PR #287, opened 2015-06-09
Add Vagrant recipe, update main `README.md` and update `.gitignore`

jeremiahsmall:
Bring over the Vagrant recipe from the Apigility skeleton app, and add some instructions to README

PR #288, opened 2015-06-09
Lite vagrantfile configuration

gianarb:
This is a PR to manage integration of basic vagrant box with shell provisioning.
Reference #287

In this moment there are a problem with the vendor directory (reference https://github.com/zendframework/ZendSkeletonApplication/pull/286#issuecomment-110294227)  because I use a little trick to decrease time of response.. We can discuss about this flow here :) 

@jeremiahsmall what do you think about this implementation?

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,15 @@ If you don't have composer installed globally then another way to create a new Z
 
 If you don't have access to curl, then install Composer into your project as per the [documentation](https://getcomposer.org/doc/00-intro.md).
 
+### Installing using Vagrant
+
+This project support a basic [Vagrant](http://docs.vagrantup.com/v2/getting-started/index.html) configuration to provide with a bash script your Skeleton Application into the VirtualBox VM.  
+
+1. Run vagrant up command
+
+   vagrant up
+
+2. Visit [http://localhost:8085](http://localhost:8085) in your browser
 
 Web Server Setup
 ----------------
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -4,33 +4,35 @@
 VAGRANTFILE_API_VERSION = '2'
 
 @script = <<SCRIPT
+DOCUMENT_ROOT_ZEND="/var/www/zf/public"
 apt-get update
-apt-get install -y apache2 git php5-cli php5 libapache2-mod-php5
+apt-get install -y apache2 git curl php5-cli php5 php5-intl libapache2-mod-php5
 echo "
 <VirtualHost *:80>
-    ServerName zf2-tutorial.localhost
-    DocumentRoot /var/www/zf/public
-    SetEnv APPLICATION_ENV "development"
-    <Directory /var/www/zf/public>
+    ServerName skeleton-zf.local
+    DocumentRoot $DOCUMENT_ROOT_ZEND
+    <Directory $DOCUMENT_ROOT_ZEND>
         DirectoryIndex index.php
         AllowOverride All
         Order allow,deny
         Allow from all
     </Directory>
 </VirtualHost>
-" > /etc/apache2/sites-available/000-default.conf
+" > /etc/apache2/sites-available/skeleton-zf.conf
 a2enmod rewrite
+a2dissite 000-default
+a2ensite skeleton-zf
 service apache2 restart
 cd /var/www/zf
-php composer.phar self-update
-COMPOSER_VENDOR_DIR=/opt/vendor php composer.phar install
-rm -rf /var/www/zf/vendor
-ln -s /opt/vendor/ /var/www/zf/vendor
+curl -Ss https://getcomposer.org/installer | php
+php composer.phar install --no-progress
+echo "** [ZEND] Visit http://localhost:8085 in your browser for to view the application **"
 SCRIPT
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.box = 'chef/ubuntu-14.04'
-  config.vm.network "forwarded_port", guest: 80, host: 8080
+  config.vm.network "forwarded_port", guest: 80, host: 8085
+  config.vm.hostname = "skeleton-zf.local"
   config.vm.synced_folder '.', '/var/www/zf'
   config.vm.provision 'shell', inline: @script
 
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,41 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+VAGRANTFILE_API_VERSION = '2'
+
+@script = <<SCRIPT
+apt-get update
+apt-get install -y apache2 git php5-cli php5 libapache2-mod-php5
+echo "
+<VirtualHost *:80>
+    ServerName zf2-tutorial.localhost
+    DocumentRoot /var/www/zf/public
+    SetEnv APPLICATION_ENV "development"
+    <Directory /var/www/zf/public>
+        DirectoryIndex index.php
+        AllowOverride All
+        Order allow,deny
+        Allow from all
+    </Directory>
+</VirtualHost>
+" > /etc/apache2/sites-available/000-default.conf
+a2enmod rewrite
+service apache2 restart
+cd /var/www/zf
+php composer.phar self-update
+COMPOSER_VENDOR_DIR=/opt/vendor php composer.phar install
+rm -rf /var/www/zf/vendor
+ln -s /opt/vendor/ /var/www/zf/vendor
+SCRIPT
+
+Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
+  config.vm.box = 'chef/ubuntu-14.04'
+  config.vm.network "forwarded_port", guest: 80, host: 8080
+  config.vm.synced_folder '.', '/var/www/zf'
+  config.vm.provision 'shell', inline: @script
+
+  config.vm.provider "virtualbox" do |vb|
+    vb.customize ["modifyvm", :id, "--memory", "512"]
+  end
+
+end
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 nbproject
 ._*
+.vagrant
 .~lock.*
 .buildpath
 .DS_Store


jeremiahsmall:
It's a very nice way to handle it. There are a few things I'd like to see tweaked though. Comments inline.

Vagrantfile
```diff
@@ -0,0 +1,41 @@
...
+        Order allow,deny
+        Allow from all
+    </Directory>
+</VirtualHost>
+" > /etc/apache2/sites-available/000-default.conf
+a2enmod rewrite
+service apache2 restart
+cd /var/www/zf
+php composer.phar self-update
+COMPOSER_VENDOR_DIR=/opt/vendor php composer.phar install
```
Maks3w: Why storage vendor outside of the project? I suspect is for the shared folder performance. Anyway put a comment explaining this.

Vagrantfile
```diff
@@ -0,0 +1,41 @@
...
+rm -rf /var/www/zf/vendor
```
Maks3w: Definitely NO!!!

This will delete forever manual vendors added in project and not commited yet!

Also the repo will be dirty (remove of tracked git files) when vagrant provision. Vagrant operations must not affect to git status.

Vagrantfile
jeremiahsmall: In my opinion, the vendor must be in the standard location relative to the project root, or else it's not available in the host machine (where my IDE is). Also, composer should run with at least `--no-progess`, so it doesn't spam the console, and also might add the rest of these switches too for nice silent install.

```
COMPOSER_VENDOR_DIR=/vagrant/vendor php composer.phar install --prefer-dist --no-plugins --no-scripts --no-progress
```

Maks3w: I prefer to display the progress if not you will perceive a cliff about what vagrant is doing along lot of minutes (think in first time run)

jeremiahsmall: --no-progress only disables the counters per repo. You still see plenty of progress. When called via cli like this, it's not able to refresh the output, so the progress counter floods the terminal with hundreds of tiny updates.

Vagrantfile
Maks3w: Just note this still been an issue even if vendor is not part of this project.

Maks3w: As noted by @jeremiahsmall remove vendor impact in the functionality of PHP IDEs.

Vagrantfile
gianarb: @Maks3w this type of implementation is difficult to read.. I agree with you but the time of response is very speedy because a lot of file (vendor) are not shared but located into the VM.. 
This is only a performance trick.. I can remove it

Vagrantfile
gianarb: @Maks3w this line of code was required because vendor was into the project.. Now I can rebase and remove it..

gianarb: No problem for me.. This "vendor trick" is only a performance improvement.. I remove it.. The response is more slow and stop :)

Vagrantfile
Maks3w: add a comment with the explanation

Vagrantfile
Maks3w: rm operations in repo folder is most than a simple code smell.

You should think how this script works the first day and the day 100th. iMay at dsay 100th custom vendor files are added and should not be removed (I need to do it in a project) Not everything is composer :)

Vagrantfile
```diff
@@ -0,0 +1,41 @@
...
+        DirectoryIndex index.php
+        AllowOverride All
+        Order allow,deny
+        Allow from all
+    </Directory>
+</VirtualHost>
+" > /etc/apache2/sites-available/000-default.conf
+a2enmod rewrite
+service apache2 restart
+cd /var/www/zf
```
Maks3w: Instead change working dir you can tell to composer what directory use as working dir

Vagrantfile
```diff
@@ -0,0 +1,41 @@
...
+VAGRANTFILE_API_VERSION = '2'
+
+@script = <<SCRIPT
+apt-get update
+apt-get install -y apache2 git php5-cli php5 libapache2-mod-php5
+echo "
+<VirtualHost *:80>
+    ServerName zf2-tutorial.localhost
+    DocumentRoot /var/www/zf/public
+    SetEnv APPLICATION_ENV "development"
```
Maks3w: Remove this setting

gianarb: I have copied this conf from Readme.md

Maks3w: I know #291 will remove it

gianarb:
Rebased

Vagrantfile
```diff
@@ -0,0 +1,38 @@
...
+# vi: set ft=ruby :
+
+VAGRANTFILE_API_VERSION = '2'
+
+@script = <<SCRIPT
+apt-get update
+apt-get install -y apache2 git curl php5-cli php5 libapache2-mod-php5
+echo "
+<VirtualHost *:80>
+    ServerName zf2-tutorial.localhost
```
Maks3w: `tutorial` word should be replace with something like `app` or `skeleton`

akrabat:
Can you add a section to README on how to use it please?

Vagrantfile
```diff
@@ -0,0 +1,38 @@
...
+    DocumentRoot /var/www/zf/public
+    <Directory /var/www/zf/public>
+        DirectoryIndex index.php
+        AllowOverride All
+        Order allow,deny
+        Allow from all
+    </Directory>
+</VirtualHost>
+" > /etc/apache2/sites-available/000-default.conf
```
Maks3w: I preffer if use his own file then enable the vhost and disable the default one.

Vagrantfile
```diff
@@ -0,0 +1,38 @@
...
+a2enmod rewrite
+service apache2 restart
+cd /var/www/zf
+curl -Ss https://getcomposer.org/installer | php
+php composer.phar install --no-progress
+SCRIPT
+
+Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
+  config.vm.box = 'chef/ubuntu-14.04'
```
Maks3w: One note of this it requires Vagrant >=1.5. Some people may be stucked on 1.4 due some plugin incompatibility.

Vagrantfile
```diff
@@ -0,0 +1,38 @@
...
+
+VAGRANTFILE_API_VERSION = '2'
+
+@script = <<SCRIPT
+apt-get update
+apt-get install -y apache2 git curl php5-cli php5 libapache2-mod-php5
+echo "
+<VirtualHost *:80>
+    ServerName zf2-tutorial.localhost
+    DocumentRoot /var/www/zf/public
```
Maks3w: Could we DRY all those repitions of `/var/www/zf/` in a variable?

Vagrantfile
```diff
@@ -0,0 +1,38 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+VAGRANTFILE_API_VERSION = '2'
+
+@script = <<SCRIPT
+apt-get update
+apt-get install -y apache2 git curl php5-cli php5 libapache2-mod-php5
```
akrabat: Do we need `php5-intl` for the `Locale` class?

Vagrantfile
Maks3w: Also extract local part fragment to a variable and use `config.vm.hostname` for to match VM hostname with the DNS

Vagrantfile
akrabat: 1.5 is fine. Vagrant is on 1.7 now… If anyone is still on 1.4, then they know what they are doing and can create their own solution.

Vagrantfile
```diff
@@ -0,0 +1,38 @@
...
+        Allow from all
+    </Directory>
+</VirtualHost>
+" > /etc/apache2/sites-available/000-default.conf
+a2enmod rewrite
+service apache2 restart
+cd /var/www/zf
+curl -Ss https://getcomposer.org/installer | php
+php composer.phar install --no-progress
+SCRIPT
```
Maks3w: Could we print an echo line with something like "Visit http://localhost:8080 in your browser for to view the application"

I feel the detail of to use port `8080` is not advised

Vagrantfile
Maks3w: Could we add `config.vm.box_url = http://opscode-vm-bento.s3.amazonaws.com/vagrant/virtualbox/opscode_ubuntu-14.04_chef-provisionerless.box` as a fallback?

Maks3w:
README.md Could include a link to http://docs.vagrantup.com/v2/getting-started/index.html for detailed instructions about virtualbox etc.

Vagrantfile
gianarb: There are other opinions about it? :+1:
Why? IMO Is the same stuff

Maks3w:
@gianarb Good work. Now I don't feel after do vagrant up some terrible thing could occur me :)

Vagrantfile
gianarb: I agree with @akrabat goodye vagrant 1.4! :)

Vagrantfile
Maks3w: Use a custom file allow to detect if the script was already executed. Thinking future changes a custom file will reduce the risks.

Vagrantfile
gianarb: @Maks3w How do you change this variable?
Why do you prefer a variable?

Vagrantfile
gianarb: I add it

Vagrantfile
Maks3w: Use a variable make simple any change

Vagrantfile
gianarb: Good idea! :)
We must choice the port :) :) 8085? boh BINGO! :tada:

gianarb:
Thanks @Maks3w for your feedback! :)

@akrabat I have rebased and maybe it's ready! :)

Vagrantfile
```diff
@@ -0,0 +1,43 @@
...
+
+Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
+  config.vm.box = 'chef/ubuntu-14.04'
+  config.vm.network "forwarded_port", guest: 80, host: 8085
+  config.vm.hostname = "skeleton-zf.local"
+  config.vm.synced_folder '.', '/var/www/zf'
+  config.vm.provision 'shell', inline: @script
+
+  config.vm.provider "virtualbox" do |vb|
+    vb.customize ["modifyvm", :id, "--memory", "512"]
```
jeremiahsmall: I'd prefer to increase the memory allocation to 1024

akrabat:
I increased memory to 1024 when merging.

gianarb:
Thanks @akrabat

README.md
```diff
@@ -36,6 +36,15 @@ If you don't have composer installed globally then another way to create a new Z
 
 If you don't have access to curl, then install Composer into your project as per the [documentation](https://getcomposer.org/doc/00-intro.md).
 
+### Installing using Vagrant
+
+This project support a basic [Vagrant](http://docs.vagrantup.com/v2/getting-started/index.html) configuration to provide with a bash script your Skeleton Application into the VirtualBox VM.
```
jeremiahsmall: This is not installation. I think this new section should go under Web Server Setup, after CLI and before Apache.

Also, suggested word-smithing:

```
### Vagrant Server

This project supports a basic [Vagrant](http://docs.vagrantup.com/v2/getting-started/index.html) configuration with an inline shell provisioner to run the Skeleton Application in a [VirtualBox](https://www.virtualbox.org/wiki/Downloads).  
```

akrabat: Good idea. - Can you PR it please?

Vagrantfile
```diff
@@ -0,0 +1,43 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+VAGRANTFILE_API_VERSION = '2'
+
+@script = <<SCRIPT
+DOCUMENT_ROOT_ZEND="/var/www/zf/public"
```
jeremiahsmall: I see the sync'd folder done as `/vagrant` quite often. What do people think about syncing `/var/www/zf` Like this? I'm not sure how I feel about it.

Vagrantfile
gianarb: You can submit a PR with a fix.. We can sync nothing and move document_root into the /vagrant dir

PR #291, opened 2015-06-09
Remove ZF1 APPLICATION_ENV

Maks3w:
APPLICATION_ENV was the ZF1 way of define the config source. ZF2 way is to have a environment config file in config folder


Maks3w:
ping @akrabat

gianarb:
:+1:  I change it in my PR @Maks3w :P

PR #293, opened 2015-06-14
Docker support

pensiero:
Added Dockerfile to allow quick setting up of a Docker container with a ready to use ZendSkeletonApplication.

Also updated README.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,10 @@
 FROM ubuntu:14.04
-MAINTAINER Oscar Fanelli <oscar.nesis@gmail.com>
+MAINTAINER Zend Framework
 ENV PROJECT_PATH /var/www/zf-skeleton
-ENV PROJECT_URL zf-skeleton.local
+ENV PROJECT_URL localhost
 ENV DEBIAN_FRONTEND noninteractive
 
 RUN apt-get update
-RUN apt-get -y --force-yes upgrade
 
 # Apache2 and PHP5
 RUN apt-get -y --force-yes install git curl apache2 php5 libapache2-mod-php5
@@ -28,9 +27,6 @@ RUN a2enconf fqdn
 # Port to expose
 EXPOSE 80
 
-# Copy site into place
-ADD . $PROJECT_PATH
-
 # VirtualHost
 RUN echo "\
 <VirtualHost *:80>\n\
@@ -47,6 +43,9 @@ RUN echo "\
 " > /etc/apache2/sites-available/zf-skeleton.conf
 RUN a2dissite 000-default
 RUN a2ensite zf-skeleton
+
+# Copy site into place
+ADD . $PROJECT_PATH
 
 # Composer
 RUN cd $PROJECT_PATH && curl -Ss https://getcomposer.org/installer | php && php composer.phar install --no-progress
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,52 @@
+FROM ubuntu:14.04
+MAINTAINER Oscar Fanelli <oscar.nesis@gmail.com>
+ENV PROJECT_PATH /var/www/zf-skeleton
+ENV PROJECT_URL zf-skeleton.local
+ENV DEBIAN_FRONTEND noninteractive
+
+RUN apt-get update
+RUN apt-get -y --force-yes upgrade
+
+# Apache2 and PHP5
+RUN apt-get -y --force-yes install git curl apache2 php5 libapache2-mod-php5
+
+# Apache2 mods
+RUN a2enmod php5
+RUN a2enmod rewrite
+
+# Apache2 environment variables
+ENV APACHE_RUN_USER www-data
+ENV APACHE_RUN_GROUP www-data
+ENV APACHE_LOG_DIR /var/log/apache2
+ENV APACHE_LOCK_DIR /var/lock/apache2
+ENV APACHE_PID_FILE /var/run/apache2.pid
+
+# Apache2 conf
+RUN echo "ServerName localhost" | sudo tee /etc/apache2/conf-available/fqdn.conf
+RUN a2enconf fqdn
+
+# Port to expose
+EXPOSE 80
+
+# Copy site into place
+ADD . $PROJECT_PATH
+
+# VirtualHost
+RUN echo "\
+<VirtualHost *:80>\n\
+  ServerName $PROJECT_URL\n\
+  DocumentRoot $PROJECT_PATH/public\n\
+  <Directory $PROJECT_PATH/>\n\
+    DirectoryIndex index.php\n\
+    Options Indexes FollowSymLinks MultiViews\n\
+    AllowOverride All\n\
+    Order deny,allow\n\
+    Allow from all\n\
+  </Directory>\n\
+</VirtualHost>\n\
+" > /etc/apache2/sites-available/zf-skeleton.conf
+RUN a2dissite 000-default
+RUN a2ensite zf-skeleton
+
+# Composer
+RUN cd $PROJECT_PATH && curl -Ss https://getcomposer.org/installer | php && php composer.phar install --no-progress
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,18 @@
 FROM ubuntu:14.04
 MAINTAINER Zend Framework
+
+# Environment variables
 ENV PROJECT_PATH /var/www/zf-skeleton
 ENV PROJECT_URL localhost
 ENV DEBIAN_FRONTEND noninteractive
 
-RUN apt-get update
-
 # Apache2 and PHP5
-RUN apt-get -y --force-yes install git curl apache2 php5 libapache2-mod-php5
+RUN apt-get update && apt-get install -y --force-yes \
+    git \
+    curl \
+    apache2 \
+    php5 \
+    libapache2-mod-php5
 
 # Apache2 mods
 RUN a2enmod php5
@@ -45,7 +50,7 @@ RUN a2dissite 000-default
 RUN a2ensite zf-skeleton
 
 # Copy site into place
-ADD . $PROJECT_PATH
+COPY . $PROJECT_PATH
 
 # Composer
 RUN cd $PROJECT_PATH && curl -Ss https://getcomposer.org/installer | php && php composer.phar install --no-progress
--- a/README.md
+++ b/README.md
@@ -77,14 +77,15 @@ This project supports a basic [Docker](https://www.docker.com/whatisdocker/) con
 4. Run `docker images` and find the *IMAGE ID* of the image
 5. Start the container (replace <image-id> with the Image ID found before):
         
-        docker run -d -p 80:80 <image-id> /usr/sbin/apache2ctl -D FOREGROUND` 
+        docker run -d -p 69:80 <image-id> /usr/sbin/apache2ctl -D FOREGROUND` 
 
-6. Go to [zf-skeleton.local](http://zf-skeleton.local) and feel the magic!
+6. Go to [localhost](http://localhost:69) and feel the magic!
 
 
 **Notes**
+- This guide map the project on the port 69, but you can change it simply replacing it in the steps *5* and *6*
 - You can change the project path editing the `PROJECT_PATH` env in the Dockerfile
-- You can change the project url editing the `PROJECT_URL` env in the Dockerfile
+- You can change the project url (used as alternative to `localhost`) editing the `PROJECT_URL` env in the Dockerfile
 
 For more informations about Dockerfile follow the [official doc](https://docs.docker.com/reference/builder/).
 
--- a/README.md
+++ b/README.md
@@ -83,11 +83,11 @@ This project supports a basic [Docker](https://www.docker.com/whatisdocker/) con
 
 
 **Notes**
-- This guide map the project on the port 69, but you can change it simply replacing it in the steps *5* and *6*
-- You can change the project path editing the `PROJECT_PATH` env in the Dockerfile
-- You can change the project url (used as alternative to `localhost`) editing the `PROJECT_URL` env in the Dockerfile
+- This recipe puts the project on the port 69, but you can choose a different port in the steps *5* and *6*
+- You can change the project path by editing `ENV PROJECT_PATH` in `Dockerfile`
+- You can change the project url by editing `ENV PROJECT_URL` in `Dockerfile`
 
-For more informations about Dockerfile follow the [official doc](https://docs.docker.com/reference/builder/).
+For more information about Dockerfile read the [official docs](https://docs.docker.com/reference/builder/).
 
 
 ### Apache setup
--- a/README.md
+++ b/README.md
@@ -57,11 +57,37 @@ This project supports a basic [Vagrant](http://docs.vagrantup.com/v2/getting-sta
 
 1. Run vagrant up command
 
-    vagrant up
+        vagrant up
+
 
 2. Visit [http://localhost:8085](http://localhost:8085) in your browser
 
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,52 @@
+FROM ubuntu:14.04
+MAINTAINER Oscar Fanelli <oscar.nesis@gmail.com>
```
Maks3w: Is this required?

Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+ENV PROJECT_PATH /var/www/zf-skeleton
+ENV PROJECT_URL zf-skeleton.local
+ENV DEBIAN_FRONTEND noninteractive
+
+RUN apt-get update
+RUN apt-get -y --force-yes upgrade
```
Maks3w: I was thinking ubuntu:14.04 is already upgraded

Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+
+# Apache2 conf
+RUN echo "ServerName localhost" | sudo tee /etc/apache2/conf-available/fqdn.conf
+RUN a2enconf fqdn
+
+# Port to expose
+EXPOSE 80
+
+# Copy site into place
+ADD . $PROJECT_PATH
```
Maks3w: I think this should be the last sentence so docker cache can reuse the maximum layers possible.

Dockerfile
pensiero: Why not? I saw that this is a common practice for those who create a Dockerfile. If it's a problem, please tell me.

README.md
```diff
@@ -57,12 +57,38 @@ This project supports a basic [Vagrant](http://docs.vagrantup.com/v2/getting-sta
 
 1. Run vagrant up command
 
-    vagrant up
+        vagrant up
```
Maks3w: Why? Instead wrap the line with ```

Dockerfile
pensiero: Right advice

README.md
```diff
@@ -57,12 +57,38 @@ This project supports a basic [Vagrant](http://docs.vagrantup.com/v2/getting-sta
...
 
+### Docker server
+
+This project supports a basic [Docker](https://www.docker.com/whatisdocker/) configuration
+
+1. [Install Docker](http://docs.docker.com/installation/)
+2. `cd` into the directory of the project
+3. Build the docker image based on the *Dockerfile*:
+        
+        docker build .
```
Maks3w: Use ``` for code blocks

README.md
```diff
@@ -57,12 +57,38 @@ This project supports a basic [Vagrant](http://docs.vagrantup.com/v2/getting-sta
...
+
+4. Run `docker images` and find the *IMAGE ID* of the image
+5. Start the container (replace <image-id> with the Image ID found before):
+        
+        docker run -d -p 80:80 <image-id> /usr/sbin/apache2ctl -D FOREGROUND`
```
Maks3w: Same as above

Dockerfile
akrabat: IMO, the maintainer should be the project. i.e. Zend Framework

Dockerfile
pensiero: Right, maybe this was necessary when specified something like `ubuntu:trusty`

README.md
```diff
@@ -57,12 +57,38 @@ This project supports a basic [Vagrant](http://docs.vagrantup.com/v2/getting-sta
...
+3. Build the docker image based on the *Dockerfile*:
+        
+        docker build .
+
+4. Run `docker images` and find the *IMAGE ID* of the image
+5. Start the container (replace <image-id> with the Image ID found before):
+        
+        docker run -d -p 80:80 <image-id> /usr/sbin/apache2ctl -D FOREGROUND` 
+
+6. Go to [zf-skeleton.local](http://zf-skeleton.local) and feel the magic!
```
Maks3w: DNS only works if host machine is updated. Can you use `localhost` here?

README.md
akrabat: I disagree. Indentation is the correct, most cross-compatible Markdown unless you need syntax highlighting.

pensiero: It was not correctly indented (and it didn't appear as code). I follow this: https://help.github.com/articles/github-flavored-markdown/#fenced-code-blocks (at the end it says: *"Keep in mind that, within lists, you must indent non-fenced code blocks eight spaces to render them properly."*)

README.md
Maks3w: Could be apache2ctl to be defined as the default entry point in Dockerfile?

README.md
Maks3w: @akrabat Code blocks boundary is the common way used in most documents, also allow to include language information.

README.md
pensiero: Unfortunately not, because the Dockerfile doesn't handle services. The only way to achieve this is with `dokku` or installing `supervisorctl`, but since this is only a "getting started application" I do not want to add too much layers (someone could love this, someone not). If a user want to start apache without writing `/usr/sbin/apache2ctl -D FOREGROUND` I think have to find out by himself.

More info: http://stackoverflow.com/questions/18047931/automatically-start-services-in-docker-container

README.md
akrabat: I strongly disagree for one-liners in README.

README.md
pensiero: I used that instead of localhost to reflect the [Vagrant configuration](https://github.com/zendframework/ZendSkeletonApplication/blob/master/Vagrantfile#L12)

README.md
Maks3w: @pensiero I don't understand. Entry point allow pass arguments

Also CMD has an explicit example for apache2ctl https://docs.docker.com/articles/dockerfile_best-practices/#cmd

README.md
akrabat: Vagrant does port mapping so that http://localhost:8888 is mapped automatically for us.

Maks3w: @pensiero If you take a look to README you will see the URL documented is localhost https://github.com/zendframework/ZendSkeletonApplication/blob/master/Vagrantfile#L29
https://github.com/zendframework/ZendSkeletonApplication/pull/293/files#diff-04c6e90faac2675aa89e2176d2eec7d8R63

README.md
pensiero: I tried, without success. If you succeed in doing it, I will add.

This is the line used:
`CMD /usr/sbin/apache2ctl -D FOREGROUND`

But when I start the container Apache2 results not started. There are [tons of similar thread](https://www.google.it/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=dockerfile+apache2+automatically) of people that do not succed in automatically start apache2 with Dockerfile.

README.md
pensiero: Right! Changing right now :)

pensiero:
@Maks3w it seems that the only way to start apache2 automatically (over installing supervisor) is to use an external script (for example `run.sh`), but this means add another file in the root of this project.

An example here: https://github.com/tutumcloud/tutum-docker-php/blob/master/run.sh#L13
And this is the added line in the Dockerfile: https://github.com/tutumcloud/tutum-docker-php/blob/master/Dockerfile#L32

If you want, I can do it, but I don't like this so much for this kind of project.

Maks3w:
This is the problem of monolithic dockerfile format

Could you create the run.sh like the vhost file was do?

If not put your right hand up and swear you won't forget what is the command

akrabat:
FWiW. I don' t want more than one file. Most users will have to delete it anyway. 

Also, I'm not yet sure that Docker is worth adding as it seems a bit more advanced to me...

pensiero:
The vhost file is created inside the container, while the run.sh file should be inside the host.
Consider that tutorials (like [this](https://www.linode.com/docs/applications/containers/linode-lamp-container-docker) explicit say *after installing it, start the apache service with...*.

Take in consideration that maybe someone could not want to start Apache2 automatically.

Maks3w:
@pensiero mmm no, run.sh is execute inside the container https://github.com/tutumcloud/tutum-docker-php/blob/master/Dockerfile#L23 :)

pensiero:
I don't know, I'm not convinced in doing it.. However, if someone say the same, I'll try to integrate.. how do you resolve this kind of decision in public projects?

README.md
jeremiahsmall: I also prefer fenced code blocks, however this file already uses indenting, so best not to mix styles.

FWIW, the syntax highlighting is a feature of [GitHub Flavored Markdown](https://help.github.com/articles/github-flavored-markdown/), so it does not work everywhere, although fenced code blocks do, as they are just a style of ordinary back tic use.

I'd rather see all the code blocks fenced, if we're going to stay with indenting, the number of leading spaces indented on the one-liners is inconsistent. Standard Markdown converts text with four spaces at the beginning of each line into a code block. For some reason we have some one-line code blocks in this file with four and some with more.

akrabat: Fenced code blocks are not in the [official spec](http://daringfireball.net/projects/markdown/); they are an extension, although a fairly common one.

If a code block is within a list, then it needs to be indented by eight spaces. First four are for the list continuation and the second four are for the code block.

README.md
```diff
@@ -57,12 +57,39 @@ This project supports a basic [Vagrant](http://docs.vagrantup.com/v2/getting-sta
...
+
+6. Go to [localhost](http://localhost:69) and feel the magic!
+
+
+**Notes**
+- This guide map the project on the port 69, but you can change it simply replacing it in the steps *5* and *6*
+- You can change the project path editing the `PROJECT_PATH` env in the Dockerfile
+- You can change the project url (used as alternative to `localhost`) editing the `PROJECT_URL` env in the Dockerfile
+
+For more informations about Dockerfile follow the [official doc](https://docs.docker.com/reference/builder/).
```
jeremiahsmall: Some edits for English style

```
**Notes**
- This recipe puts the project on the port 69, but you can choose a different port it in the steps *5* and *6*
- You can change the project path by editing `ENV PROJECT_PATH` in `Dockerfile`
- You can change the project url by editing `ENV PROJECT_URL` in `Dockerfile`

For more information about Dockerfile read the [official docs](https://docs.docker.com/reference/builder/).
```

pensiero: Thanks a lot @jeremiahsmall :+1:

README.md
jeremiahsmall: Oh yes, I see it's consistently 8 inside if lists and 4 outside of lists.

jeremiahsmall:
> Also, I'm not yet sure that Docker is worth adding as it seems a bit more advanced to me...

@akrabat, personally I don't think it's going to confuse anyone, as they can just ignore it if they prefer to, but it is a nice addition.

akrabat:
As long as it's just one file, I'm tending to agree. Undoing all the translation stuff is more of a hassle than deleting a single file :)

gianarb:
We can migrate between provisioning tool for all life :+1:

gianarb:
@weierophinney what do you think about this PR? :)
Docker is a very important tool and maybe a good dockerfile to create a ZF app container is a good feature..
Can we register it into the official repository after merge? :)

jeremiahsmall:
+1 @gianarb for adding to registry.

akrabat:
Has anyone other than @pensiero tested this and proved that it works for development purposes?

svycka:
I don't think it's worth adding, unless this is for minimal demo demonstration that it works, but we have vagrant for this. If someone is planning using this for development with docker or vagrant they will need a lot more like xdebug, remote debugger configured, profiller, databases and more. This setup is not so easy to edit and maintain. So everyone who needs this will anyway create their own setup. For a good setup it wont be one file...

Thinkscape:
@svycka Not entirely. When you look at [current Vagrantfile](https://github.com/zendframework/ZendSkeletonApplication/blob/master/Vagrantfile) it is nothing more than an Ubuntu + php + apache2. It's got no devel tools, nothing fancy. It's a baseline, simple, example provisioning just to get one started.

The proposed dockerfile serves similar purpose - it is an EXAMPLE dockerfile, showing the minimal configuration and provisioning to get it going. Anything beyond that is outside of the scope of this repository.

After it's merged I don't see any reason not to submit it to docker hub.

svycka:
@Thinkscape  I didn't said that vagrant setup is good. Apigility uses https://puphpet.com/ for this and this is a lot better. With https://github.com/zfcampus/zf-apigility-skeleton/pull/106 will be even better, also will be not so much overhead compared to flexibility.

pensiero:
Guys, I really can't understand why so much problems for a basilar configuration file. ZendSkeletonApplication is not a *real* application, is an example for developers that approach with Zend Framework. And when it happens, the developer could probably need to set up an environment with Vagrant, Docker and so on, as it happened to me and my colleagues.

For this reason, I think that it's nice to provide the most important configuration files that help them in the setting up. I don't think that 1 configuration file is worth adding if it can help even only 1 person in approaching to Zend Framework more easily.

svycka:
I just think that many(more) people use this as starting point and not as of example. And as starting point it is very handy to have easy configurable development envirovement, but current setups does not provide this. Thats all.

Thinkscape:
Both your argumentation and what @jeremiahsmall mentions on zfcampus/zf-apigility-skeleton#106 are debatable. For example, should the VM include MySQL ? Why not postgresql ? No love for mongodb ? Why does it not go with FPM ? Why this particular PHP version and not version X.Y ? etc.

"Puphet" and "Phansible" are great examples of a real-world solution to GUI-driven Vagrant + provisioning generation which anyone can use. There is no need to try to replicate any of that or make a decision for the majority of users which settings we would go with. The proposed Docker and Vagrant files are meant to be "minimal" and "educational", not be a real-world solutions. As @pensiero mentioned, the whole repository is meant as an example, a learning material for newcomers. It is not a real application. It's barely a suggestion on best practices.  

Quote from [README.md](https://github.com/zendframework/ZendSkeletonApplication/blob/master/README.md):

> This is a simple, skeleton application using the ZF2 MVC layer and module systems. This application is meant to be used as a starting place for those looking to get their feet wet with ZF2.

jeremiahsmall:
The Vagrant recipe in  zfcampus/zf-apigility-skeleton#106 was created with Puphpet, and I originally proposed adding the identical Vagrant recipe to this project as well, but there was objection to the weight of it. That's why the light weight version was created in this app.

Now having tried the light weight version that was accepted and I had a chance to play with it, my first issue with it is that I find it to be just too bare-bones. @Thinkscape is right that it's debatable what to put in the example, but I think having no database provisioned at all is a bit silly. If you want to add you own features, you basically must dig deep into Vagrant and start to learn how to do custom provisioning, or you just have to throw it away, and replace it with something else. 

I'm in favor of providing a Puphpet Vagrant recipe, as that is very easy to customize by just editing `puphpet/config.yaml` (if you want other database, etc). As long as we're ignoring the modules dir (as in my recent open proposal), it's not messy, which was the original reason it was abandoned in this skeleton.

This brings me to my second issue: lack of parity between this skeleton and the Apigility skeleton. I think that the same recipe(s) should be provided in both skeletons.

As far as maintaining Docker support, I really don't care if we have it. Personally I think Vagrant is a more useful way to get running on a local dev machine. That said if we do keep it, I don't mind, but I'd prefer to see parity between the skeleton apps with Docker too.

pensiero:
Guys, sorry if I tell you but you are really spending to much time discussing in something that, if integrated, will not overturn the life of anybody, it's just an quick help for someone.

If we discuss on everything in this way, it's impossible to move on.

pensiero:
I opened this PR about 1 year ago. Docker support grows everywhere, is someone interested in merging this?

weierophinney:
Accomplished with #329 (to release soon).

pensiero:
👍 (the PRs number are specular, so perfect!)

PR #305, opened 2015-09-08
Zend Framework auto installation file

Stanimirdim92:
Hi,

I have created an installation script for Zend Framework. All you have to do is click on the "Start installation" button and the script will install composer (this means that you no longer will need to have composer installed or use a terminal) and download Zend Framework for you. You will need to store composer.phrar in the root directory in order to install composer.
After the installation of the vendor folder, the user will be asked to setup a database with all the information like user, port, ip etc. All the information will be stored in local.php file.

The script also simplifies init_autoloader.php file.

Currently there is no design for messages or forms, but that's easly fixable and it will be done soon.


Stanimirdim92:
@Ocramius Do you think that this can be merged? We can even remove init_autoloader.php and move $loader to index.php to simplify things.

Ocramius:
I didn't see this before, sorry.

I personally don't see a need for functionality like this, as it poses more of a security issue than an actual functionality.
I'd prefer people to just learn to use the tool or use a CLI binary rather than treat the tool like a self-installing magic box.

That's my take on it at least.

gianarb:
I agree with @Ocramius

vrkansagara:
:+1:

Stanimirdim92:
Closed. There is zend-expressive which gives you more flexible solution.

PR #306, opened 2015-09-11
remove ZF2_PATH env usage as requirement is zf ~2.5

samsonasik:


gianarb:
:+1:

Farsideman:
Thank you for your comments. I have implemented the changes but this has no effect on my problem. 

As I am rather new to the framework, perhaps I can provide more detail.

Platform: MAMP Pro 3.2.1

This platform worked with zf2 version 2.4 and the Album project described in this version of the  documentation.

Then Composer pushed the zf2 version 2.5 update. The Album project broke. Is there 2.5 documentation? Is there a revised album project? I cloned the latest album code from the git repository and it didn't work either. I then changed the init_autoloader.php file as suggested with no change.

The problem with the Album project is as follows:

The error reporting will in development does not work. A Undefined index: APPLICATION_ENV is encountered.

The index action works. The add, edit, and delete action update the database but returns a blank page instead of the updated list of albums.

Any assistance would be appreciated.

Thank you.

samsonasik:
APPLICATION_ENV seems not related with ZF2_PATH env, you can define the APPLICATION_ENV in you vhost like in documenation: http://zf2.readthedocs.org/en/latest/user-guide/skeleton-application.html

samsonasik:
anyway, this change should for zf ~2.5, that's why i use titlle : "~2.5"

Farsideman:
Thank you, the APPLICATION_ENV problem is solved. I hope someone can comment on the other issue.

Farsideman:
A little more information:
In the version 2.4 user guide on the Forms and Actions page under Adding New Albums the following code is described:

1 // Redirect to list of albums
2 return $this->redirect()->toRoute('album');

After we have saved the new album row, we redirect back to the list of albums using the Redirect controller plugin.

1 return array('form' => $form);

Finally, we return the variables that we want assigned to the view. In this case, just the form object. Note that Zend Framework 2 also allows you to simply return an array containing the variables to be assigned to the view and it will create a ViewModel behind the scenes for you. This saves a little typing.

This seems to be the section that fails and returns a blank page. It is now making it to the view. How would you debug this?

samsonasik:
I think that is separate issue, or more a _question_ instead of issue. Not related with this PR

Warm regards,

Abdul Malik Ikhsan

Pada 15 Sep 2015, pukul 04.16, John Pedosuk <notifications@github.com> menulis:

Farsideman:
Thanks,

I will post it as a separate issue.

weierophinney:
Accomplished with #329 ; thanks for the nudge, @samsonasik !

PR #315, opened 2015-10-08
Update index.php

NeftaliYagua:
This check should not be optional,
The getenv function prevents a "NOTICE" for undefined variable APPLICATION_ENV.

### DIFF
--- a/public/index.php
+++ b/public/index.php
@@ -1,4 +1,11 @@
 <?php
+ /**
+  * Display all errors when APPLICATION_ENV is development.
+  */
+if (getenv('APPLICATION_ENV') && getenv('APPLICATION_ENV') == 'development') {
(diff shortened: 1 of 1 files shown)


public/index.php
```diff
@@ -1,4 +1,11 @@
 <?php
+ /**
+  * Display all errors when APPLICATION_ENV is development.
+  */
+if (getenv('APPLICATION_ENV') && getenv('APPLICATION_ENV') == 'development') {
```
Maks3w: The first getenv is not necessary

NeftaliYagua: You are right!

gianarb:
ZF community doesn't subsidize this approach, we use the local configuration file to change the environment.

Thanks

vrkansagara:
:+1:

Issue #313, opened 2015-09-28
Kill non-composer installation methods?


Ocramius:
About time, no?

RalfEggert:
Yep! Makes everything easier for beginners.

gianarb:
:+1:

Stanimirdim92:
:+1:

gianarb:
How can we proceed? I'm ready to delete all lines :)

Stanimirdim92:
Any chance to remove init_autoloader and move $loader in index.php? @Ocramius

Ocramius:
That's the idea.

Stanimirdim92:
Do it, pretty please.

Moln:
:+1:

Thinkscape:
Yup
On Tue, May 17, 2016 at 7:20 AM, Moln <notifications@github.com> wrote:

> [image: :+1:]
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/zendframework/ZendSkeletonApplication/issues/313#issuecomment-219621885>
>
-- 
Artur Bodera
abodera@gmail.com
http://thinkscape.pro

weierophinney:
Accomplished with #329 (still to be released).

PR #316, opened 2015-10-18
Drop supports for Zend\Loader

gianarb:
Reference with #313.
Use composer to manage Zend Framework Application autoloader.


public/index.php
```diff
@@ -5,17 +5,13 @@
...
+// Composer autoloading
+if (!file_exists('vendor/autoload.php')) {
+    throw new RuntimeException(
+        'Unable to load ZF2. Run `php composer.phar install` or define a ZF2_PATH environment variable.'
+    );
 }
-
-// Setup autoloading
-require 'init_autoloader.php';
+$loader = include 'vendor/autoload.php';
```
danizord: No need to store the result in a variable

danizord: Should use `require` instead of `include`

public/index.php
```diff
@@ -5,17 +5,13 @@
...
-if (php_sapi_name() === 'cli-server') {
-    $path = realpath(__DIR__ . parse_url($_SERVER['REQUEST_URI'], PHP_URL_PATH));
-    if (__FILE__ !== $path && is_file($path)) {
-        return false;
-    }
-    unset($path);
+// Composer autoloading
+if (!file_exists('vendor/autoload.php')) {
+    throw new RuntimeException(
+        'Unable to load ZF2. Run `php composer.phar install` or define a ZF2_PATH environment variable.'
```
danizord: `ZF2_PATH` env variable will not work anymore

composer.json
```diff
@@ -10,5 +10,10 @@
     "require": {
         "php": ">=5.5",
         "zendframework/zendframework": "~2.5"
+    },
+    "autoload": {
+        "psr-4": {
+          "Application\\": "module/Application/src/Application"
```
adamlundrigan: The module is set up as PSR-0 currently.  Could we drop the `Application` folder inside `src` to keep with the intent of PSR-4, or if that's not on then change this to PSR-0? 

And, yes, that's probably the most pedantic programming-related statement I've ever written.

gianarb: Sure.. I replace with psr-0.. :) 

@adamlundrigan don't worry :) You are here to help me and I'm very happy to maintain this record! :trophy:

public/index.php
```diff
@@ -14,8 +14,14 @@
...
-// Setup autoloading
-require 'init_autoloader.php';
+// Composer autoloading
+if (!file_exists('vendor/autoload.php')) {
+    throw new RuntimeException(
+        'Unable to load ZF2. Run `php composer.phar install`.'
+    );
+}
+
+require 'vendor/autoload.php';
```
Ocramius: `require_once` please

public/index.php
```diff
@@ -14,8 +14,14 @@
     unset($path);
 }
 
-// Setup autoloading
-require 'init_autoloader.php';
+// Composer autoloading
+if (!file_exists('vendor/autoload.php')) {
```
Ocramius: This `if` block can actually be skipped, as it causes a stat call. Not sure if it would make the FW actually harder to use tho :-(

composer.json
```diff
@@ -10,5 +10,10 @@
     "require": {
         "php": ">=5.5",
         "zendframework/zendframework": "~2.5"
+    },
+    "autoload": {
+        "psr-0": {
```
Ocramius: You can use `psr-4` here, by simply configuring `"Application\\": "module/Application/src/Application"`

module/Application/Module.php
```diff
@@ -25,15 +25,4 @@ public function getConfig()
...
-    public function getAutoloaderConfig()
-    {
-        return array(
-            'Zend\Loader\StandardAutoloader' => array(
-                'namespaces' => array(
-                    __NAMESPACE__ => __DIR__ . '/src/' . __NAMESPACE__,
-                ),
-            ),
-        );
-    }
```
Ocramius: Could you move the `Module.php` to the correct location?

gianarb: The correct location is into the `module/Application/src/Application`?

composer.json
Maks3w: Previous was correct.

adamlundrigan: It works but the second Application folder is superfluous with PSR-4 so why have it?  Like I said: extreme pedantry.

Stanimirdim92: No need for the second Application folder in the path. Jut use "Application\\": "module/Application/src" for psr-4

danizord: I'm :+1: to move the contents of `module/Application/src/Application` down to `module/Application/src`.

gianarb:
Guys.. News? @Ocramius

module/Application/src/Application/Module.php
```diff
@@ -23,17 +23,6 @@ public function onBootstrap(MvcEvent $e)
...
-    public function getAutoloaderConfig()
-    {
-        return array(
-            'Zend\Loader\StandardAutoloader' => array(
-                'namespaces' => array(
-                    __NAMESPACE__ => __DIR__ . '/src/' . __NAMESPACE__,
-                ),
-            ),
-        );
+        return include __DIR__ . '/../../config/module.config.php';
```
danizord: `require`

gianarb: :confused: Thanks!

gianarb: Can you help me to understand? :) Thanks
https://github.com/doctrine/DoctrineModule/blob/master/src/DoctrineModule/Module.php#L74

danizord: @gianarb the config file is **required** here. If config file is not found, then it should raise a fatal error and stop the script execution.

samsonasik:
I prefer `init_autoloader.php` still there, but with simpler logic depends on composer:

```php
use Zend\Loader\AutoloaderFactory;

// Composer autoloading
if (file_exists('vendor/autoload.php')) {
    $loader = include 'vendor/autoload.php';
}

if (class_exists(AutoloaderFactory::class)) {
    return;
}

throw new RuntimeException('Unable to load ZF2. Run `php composer.phar install`');
```

gianarb:
@weierophinney do you have some news? 
We already use this kind of approach for zf-web
https://github.com/zendframework/zf-web/blob/master/composer.json

weierophinney:
Accomplished with #329 - thanks for the nudge, @gianarb !

PR #317, opened 2015-10-21
Add ./serve.sh quick runing zf2 app with php -S

netroby:
Signed-off-by: ZhiFeng Hu <hufeng1987@gmail.com>


gianarb:
I see this PR unnecessary.
maybe we can add a command in ZFTool

vrkansagara:
Its good to be in separate file. :+1: @netroby 
But this is already given into the quick start guide so no need to make extra file for that.

PR #327, opened 2016-04-07
Fix language metadata

theboolean:
Fix: it_IT language had metadata related to german translation


weierophinney:
We've removed translations for the upcoming v3 release; thank you for your efforts, however!

PR #329, opened 2016-04-18
Prepare for zend-mvc v3

weierophinney:
For the v3 release, we will be pulling in a smaller set of components out-of-the-box. These are primarily defined by zend-mvc, but also include:

- ~~zend-config (used by the zend-modulemanager `ConfigListener` by default)~~ (zend-modulemanager now defines the zend-config requirement)
- [zend-skeleton-installer](https://zendframework.github.io/zend-skeleton-installer/) for prompting for optional packages/capabilities.
- [zend-component-installer](https://zendframework.github.io/zend-component-installer), to automate registration of components as modules.

Additionally, we plan to leverage Composer for all autoloading, which means:

- we can register a PSR-4 entry for the `Application` module, and remove one directory level;
- the `Application` module no longer needs to setup autoloading.

For a minimal application, we do not want or need internationalization. As such, this patch also *removes all translations* and calls to translate strings in templates. We can explore re-adding them later, or offering an alternative skeleton application with more default features (including i18n, forms, databases, etc.).

At the time of submission, this patch provides a working skeleton with the current zend-mvc develop branch.

## Testing

If you wish to test:

- Clone the repository locally
- Enter the cloned repository
- Switch to the develop branch: `git checkout -b develop origin/develop`
- Apply the changes: `wget https://patch-diff.githubusercontent.com/raw/zendframework/ZendSkeletonApplication/pull/329.patch && git am 329.patch` (or, if you use the hub command: `git merge https://github.com/zendframework/ZendSkeletonApplication/pull/329`)
- Execute `composer install`

If you want to use vagrant:

- Start it: `vagrant up`
- Run composer within it: `vagrant ssh -c 'cd /var/www ; composer install'`
- Start testing

If you want to use docker:

- Build and run it: `docker-compose up -d --build`
- Run composer within it: `docker-compose run zf composer install`
- Start testing.

Please report any issues you notice or provide constructive feedback via comments to this pull request, or issue a pull request against my branch.

### Installer

- [x] Prompt for caching
  - will install zend-cache
- [x] Prompt for console
  - prompt should indicate users should migrate to zf-console
  - will install zend-mvc-console
- [x] Prompt for database
  - will install zend-db
- [x] Prompt for forms
  - will install zend-form
- [x] Prompt for i18n
  - will install zend-mvc-i18n
- [x] Prompt for JSON
  - will install zend-json
- [x] Prompt for logging
  - will install zend-log
- [x] Prompt for PSR-7 middleware dispatcher
  - will install zend-psr7bridge
- [x] Prompt for sessions
  - will install zend-session
- [x] Prompt for testing facilities
  - will install zend-test *as dev requirement*
- [x] Prompt for plugins
  - will install zend-mvc-plugin-*
- [x] Prompt for zend-di integration
  - will install zend-servicemanager-di

In each case, this should register the module(s) with the application, so that the user does not need to be prompted twice.

### Documentation

- ModuleRouteListener is removed from the skeleton. This won't affect existing users, but *will* affect experienced users who originally relied on it being active in new skeleton projects.
- The `/[:controller][/:action]]` route was removed from the skeleton. Again, it will not affect existing users, but *will* affect experienced users who originally relied on it being active in new skeleton projects.

These items have been added to the [maintainers wiki](https://github.com/zendframework/maintainers/wiki/Tutorials-and-documentation-tasks-for-v3).

## Updates

- *2016-05-17: added TODO list tracking installer and documentation tasks remaining.*

Close #293


weierophinney:
This is a WIP; do not merge until zend-mvc is tagged and/or zendframework is tagged!

init_autoloader.php
```diff
@@ -1,52 +1,13 @@
...
-                'autoregister_zf' => true
-            )
-        ));
-    }
-}
+include 'vendor/autoload.php';
 
-if (!class_exists('Zend\Loader\AutoloaderFactory')) {
-    throw new RuntimeException('Unable to load ZF2. Run `php composer.phar install` or define a ZF2_PATH environment variable.');
+if (! class_exists('Zend\Mvc\Application')) {
```
Ocramius: `::class`

module/Application/view/error/404.phtml
```diff
@@ -31,14 +31,14 @@ switch ($this->reason) {
...
-    <dt><?php echo $this->translate('Controller') ?>:</dt>
+    <dt>Controller:</dt>
     <dd><?php echo $this->escapeHtml($this->controller) ?>
 <?php
 if (isset($this->controller_class)
     && $this->controller_class
     && $this->controller_class != $this->controller
 ) {
-    echo '(' . sprintf($this->translate('resolves to %s'), $this->escapeHtml($this->controller_class)) . ')';
+    echo '(' . sprintf('resolves to %s', $this->escapeHtml($this->controller_class)) . ')';
```
Ocramius: Did we completely remove the translation support?

module/Application/src/Module.php
```diff
@@ -0,0 +1,27 @@
...
+use Zend\Mvc\ModuleRouteListener;
+use Zend\Mvc\MvcEvent;
+
+class Module
+{
+    const VERSION = '3.0.0dev';
+
+    public function onBootstrap(MvcEvent $e)
+    {
+        $moduleRouteListener = new ModuleRouteListener();
```
Ocramius: Wondering if this is still needed

config/application.config.php
```diff
@@ -8,6 +8,8 @@
 return array(
     // This should be an array of module namespaces used in the application.
     'modules' => array(
+        'Zend\Router',
```
Ocramius: IIRC, we can reference module classes also via their FQCN

Ocramius:
A few notes here:

 * not sure about translations, although they did make the skeleton MUCH fatter for very little gain. Including a translation module would be sufficient. Keeping the `->translate()` calls would also be sufficient, for now.
 * I can't find direct references to `Zend\Config` in this code. Is the `zend-modulemanager` using `class_exists()`? If so, we should probably make this particular dependency usage explicit.
 * `zend-component-installer` I'm trying that nao \o/

Ocramius:
A few other notes:

 * `Zend\Loader` still seems to be required: what does it take to chop its head off?
 * `Zend\Http` also required, but I see no diactoros: does that mean that PSR-7 support won't be there OOTB?
 * aren't required components automatically injected in `application.config.php` by the zendframework/zend-component-installer`? I glanced over the code at https://github.com/zendframework/zend-component-installer/blob/07fa730b11041d48106f658b655230828dc76ff1/src/ComponentInstaller.php#L187 - seems like that's the case

Ocramius:
Confirmed: adding them to `application.config.php` is probably not needed, and it would expose the users to the process of configuring modules (possibly good):

ecf53c

back-2-95:
Very nice, it looks very clear and I like the way other zend-components are registered as modules. Removing one directory level is also very nice. We developers are lazy and like every small thing which makes our lives easier.

For the skeleton, I would convert all array(...) to use short array syntax. Old arrays exists on config files and in the layout.phtml.

Thanks for good work guys!

module/Application/view/application/index/index.phtml
```diff
@@ -1,43 +1,43 @@
...
             </div>
             <div class="panel-body">
-                <p><?php echo sprintf($this->translate('If you need any help or support while developing with ZF2, you may reach us via IRC: %s#zftalk on Freenode%s. We\'d love to hear any questions or feedback you may have regarding the beta releases. Alternatively, you may subscribe and post questions to the %smailing lists%s.'), '<a href="irc://irc.freenode.net/zftalk">', '</a>', '<a href="http://framework.zend.com/wiki/display/ZFDEV/Mailing+Lists">', '</a>') ?></p>
-                <p><a class="btn btn-success pull-right" href="http://webchat.freenode.net?channels=zftalk" target="_blank"><?php echo $this->translate('Ping us on IRC') ?> &raquo;</a></p>
+                <p><?php echo sprintf('If you need any help or support while developing with ZF2, you may reach us via IRC: %s#zftalk on Freenode%s. We\'d love to hear any questions or feedback you may have regarding the beta releases. Alternatively, you may subscribe and post questions to the %smailing lists%s.', '<a href="irc://irc.freenode.net/zftalk">', '</a>', '<a href="http://framework.zend.com/wiki/display/ZFDEV/Mailing+Lists">', '</a>') ?></p>
```
samsonasik: s/beta release/this release

samsonasik: s/beta releases/this release

init_autoloader.php
Stanimirdim92: Why not move that code in index.php? Imo it will simplify things a little more.

init_autoloader.php
```diff
@@ -1,52 +1,13 @@
...
-    } else {
-        include $zf2Path . '/Zend/Loader/AutoloaderFactory.php';
-        Zend\Loader\AutoloaderFactory::factory(array(
-            'Zend\Loader\StandardAutoloader' => array(
-                'autoregister_zf' => true
-            )
-        ));
-    }
-}
+include 'vendor/autoload.php';
```
samsonasik: as no file_exists check, I think `@` prefix ( `@include` usage ) should be provided as the error needs to be shown to user is the RuntimeException error

samsonasik:
@weierophinney I created PR https://github.com/weierophinney/ZendSkeletonApplication/pull/3 to your branch for some feedbacks above ( wording, ::class, short array syntax )

module/Application/view/error/404.phtml
weierophinney: For the skeleton, yes; see the issue summary. :wink:

(Plan is to potentially ship multiple skeletons, or use an installer to provide different features.)

module/Application/src/Module.php
weierophinney: I have no clue how often it's used in real-world applications, unfortunately. I suspect it could be useful for having a root route that defines a namespace, but cannot match without child routes, but even then, I personally would rather use FQCN for the controllers anyways.

Definitely up for debate.

config/application.config.php
weierophinney: Not currently. The `ModuleResolverListener` always appends `Module` to the module name in order to determine if the module exists.

module/Application/src/Module.php
Ocramius: Can we get rid of it? This is so magic that I can barely even describe it to a newcomer :P

module/Application/src/Module.php
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+namespace Application;
+
+use Zend\Mvc\ModuleRouteListener;
+use Zend\Mvc\MvcEvent;
+
+class Module
+{
+    const VERSION = '3.0.0dev';
```
Ocramius: This is shameless advertising, but `ocramius/package-versions` deals with it, eventually :)

weierophinney:
@Ocramius — some responses:

> - not sure about translations, although they did make the skeleton MUCH fatter for very little gain. Including a translation module would be sufficient. Keeping the ->translate() calls would also be sufficient, for now.

The `translate()` helper is part of zend-i18n, and would thus require installing zend-mvc-i18n as a dependency (which also brings in zend-i18n). We can certainly do that, but my thought is to have:

- a base, minimal skeleton, with only the bare minimum necessary to create an MVC application, and then potentially add skeletons with more features.
- an installer, ala zend-expressive, that allows you to opt-in to features:
  - "Do you want the console integration?"
  - "Do you want i18n integration?"
  - "Do you want to use zend-db?"
  - "Do you need caching?"
  - etc.

One consistent complaint I've heard is from developers who start with the skeleton, and then need to strip out a whole bunch of stuff they don't need. If we have a minimal base on which we can expand, we solve that problem.

> - I can't find direct references to Zend\Config in this code. Is the zend-modulemanager using class_exists()? If so, we should probably make this particular dependency usage explicit.

It's used by the `ConfigListener` of zend-modulemanager, which I discovered that when testing out the skeleton. At this point, I'm unsure if the dependency should go in:

- zend-modulemanager (`ConfigListener` is technically optional, though)
- zend-mvc (as it enables the `ConfigListener` by default)
- the skeleton application

Thoughts?

> - zend-component-installer I'm trying that nao \o/

Interestingly, the `Zend\Router` and `Zend\Validator` entries in the application configuration *were added by the component installer* when I first did a `composer install`. :smile: (I'm *very* excited by this!)

weierophinney:
@Ocramius — more responses:

- Zend\Loader still seems to be required: what does it take to chop its head off?

It's currently required by zend-modulemanager, for autoloading modules that do not have autoloading setup. (See `Zend\ModuleManager\Listener\ModuleLoaderListener`.)

Since one goal is to have a clear and clean upgrade path, we currently need to keep that. What we *could* do, however, is deprecate the usage; the question is if we should do that in a 2.x series of zend-modulemanager, or in the v3 release.

If we do it in the 2.x series, we could emit a deprecation notice from `Zend\ModuleManager\Listener\AutoloaderListener` whenever it receives autoloader configuration, prompting the user to add autoloading configuration to their `composer.json`; we could even link to a documentation page detailing the various approaches.

Thoughts?

- Zend\Http also required, but I see no diactoros: does that mean that PSR-7 support won't be there OOTB?

Out of the box, no, not for the minimal skeleton. zend-psr7bridge *is* listed as a suggested requirement for zend-mvc already, however, and the suggestion text indicates it should be installed if you plan to use PSR-7 middleware within your application.

- aren't required components automatically injected in application.config.php by the zendframework/zend-component-installer`? I glanced over the code at https://github.com/zendframework/zend-component-installer/blob/07fa730b11041d48106f658b655230828dc76ff1/src/ComponentInstaller.php#L187 - seems like that's the case

Yes and no. zend-component-installer *prompts* you to add the component, but gives you the option of saying "no" (as well as saying, "remember my selection for all remaining components on this installation run). It also checks to see if the component *is already listed*, and, if so, will not prompt. My feeling is that components *required* by the base skeleton should be listed in the configuration as well, so that you only get prompted the first time you add a new component.

weierophinney:
> For the skeleton, I would convert all array(...) to use short array syntax. Old arrays exists on config files and in the layout.phtml.

Noted; can't believe I missed that on the first pass through!

init_autoloader.php
weierophinney: PHP will barf all over the place at this point anyways, and you'll get both an `E_WARNING` and the following `RuntimeException`. Essentially, I don't want to suppress errors or do additional error checking for something that represents a fundamentally broken install anyways.

module/Application/src/Module.php
weierophinney: Oh, that's slick!

/me goes to add that in the next push...

weierophinney: DAMMIT, @Ocramius ... PHP 7 as the minimum requirement?!?!?! (now I understand the "eventually" comment, though...)

Ocramius:
> The `translate()` helper is part of zend-i18n, and would thus require installing zend-mvc-i18n as a dependency (which also brings in zend-i18n). We can certainly do that, but my thought is to have:
 > 
 > - a base, minimal skeleton, with only the bare minimum necessary to create an MVC application, and then potentially add skeletons with more features.
 > - an installer, ala zend-expressive, that allows you to opt-in to features:
 > - "Do you want the console integration?"
 >   - "Do you want i18n integration?"
 >   - "Do you want to use zend-db?"
 >   - "Do you need caching?"
 >   - etc.

Agreed, let's keep i18n out then.

 > > - I can't find direct references to Zend\Config in this code. Is the zend-modulemanager using class_exists()? If so, we should probably make this particular dependency usage explicit.
 > 
 > It's used by the `ConfigListener` of zend-modulemanager, which I discovered that when testing out the skeleton. At this point, I'm unsure if the dependency should go in:

It should probably live in the `module-manager`. The 95% use-case of the module manager is merging config.

 > Out of the box, no, not for the minimal skeleton. zend-psr7bridge *is* listed as a suggested requirement for zend-mvc already, however, and the suggestion text indicates it should be installed if you plan to use PSR-7 middleware within your application.

Fair enough. I feel like we should push further on this in future, but maybe not now.

 > Yes and no. zend-component-installer *prompts* you to add the component, but gives you the option of saying "no" (as well as saying, "remember my selection for all remaining components on this installation run). It also checks to see if the component *is already listed*, and, if so, will not prompt. My feeling is that components *required* by the base skeleton should be listed in the configuration as well, so that you only get prompted the first time you add a new component.

My thought here is that while the user may answer "no", he may actually learn that he will be prompted. Not having that is also fine, IMO, but we're committing post-install artifacts. Still: good for usability, let's keep it as-is, I'd say.

module/Application/src/Module.php
Ocramius: Hah, totally forgot about that, sorry :-) 

Yeah, next time then :-P

Thinkscape:
> Out of the box, no, not for the minimal skeleton. zend-psr7bridge is listed as a suggested requirement for zend-mvc already, however, and the suggestion text indicates it should be installed if you plan to use PSR-7 middleware within your application.
> Fair enough. I feel like we should push further on this in future, but maybe not now.

Would it be that crazy if the "installer" and skeleton was merged with [zend-expressive-skeleton](https://github.com/zendframework/zend-expressive-skeleton) ?

#### If it was, there are possible cons:
* People would still need to decide whether to go PSR7 or "legacy mvc", requires understanding what it is.
* Increases package complexity slightly.

#### Pros:
* Advertising: 3.0.0 equals the *future version* - expressive becomes integral part of new version.
* Only one skeleton to maintain instead of two.
* People actually start using PSR7 when they "default to ZF" or just want to try "the new ZF". 
* Because there's still a choice between PSR7 and "legacy mvc" in the installer, people can choose the latter and have somewhat easier migration path from ZF2.
* Less legacy stuff. No need to wait for ZF4 or ZF3.1 or whatever else that cuts the cord.

Current blog posts and available docs do not clearly explain the PSR7 transition IMO. This would effectively solve the problem as ZF3 skeleton would give one clear choice. Through the "power of defaults" (default = expressive), the installer would be educating and promoting PSR7.

samsonasik:
I think for 'controllers' -> 'invokables' config, we can use factories and use `Zend\ServiceManager\Factory\InvokableFactory` ? and change `Application\Controller\Index` to `Application\Controller\IndexController` for the key.

weierophinney:
@Thinkscape —

> Would it be that crazy if the "installer" and skeleton was merged with zend-expressive-skeleton ?

Yes, yes it would. :smile:

Here's the thing: zend-mvc and zend-expressive represent very different paradigms, including configuration and workflow. Merging the two would be akin to violating the SRP.

While I appreciate the pros and cons list you've created, I think you vastly underestimate the point about package complexity, and user comprehension. Merging the two installers is difficult for a number of reasons:

- Right now, we can have the skeleton reflect 90% of the application, with the last 10% being from installation artifacts created by the installer. Merging the two would switch that ratio around and require generation of around 90% of the skeleton by the installer.
- This in turn makes maintenance far harder for the bulk of changes. Many changes we make currently can be made in files *in their final location*; moving this all into installation artifacts makes understanding where the changes need to be made more difficult.
- End users *using* the installer now need to know the difference between zend-mvc and zend-expressive in order to make a decision as to what to do next. This is not at all trivial, and while we could default to one or the other (likely Expressive), I suspect that this would be a huge hurdle.

At this point, we've trimmed the cruft from the MVC, and, believe it or not, I think we've made it a far more viable option as a full-stack framework. For those who want that kind of integration, it's a good option. But for those who want to use PSR-7 and middleware, or who want fine-grained control over what they install and use in their application, Expressive will be a better fit. *They are different choices, targeting different sets of developers.*

The takeaway is: we need to figure out our marketing messages around the two, and make sure there's readily accessible information so developers can make the correct choice for their project. But part of that choice is figuring out *which* project to start with, and merging those complicates that choice.

weierophinney:
@samsonasik Thanks for the patches!

module/Application/src/Module.php
akrabat: FWiW, I agree with @Ocramius on this one.

Thinkscape:
There's two major concerns I see.

Even with extremely well constructed, immaculate PR play you still end up having to explain "Zend Framework" (the parent `github.com/zendframework` group and project marketing name). What is it then? 

If you follow the "bag of loosely-coupled coupled components" rhetoric, then you're allowed to give out several (competing) implementations of say, MVC. Thanks to interop efforts, it now goes even beyond `github.com/zendframework` and one can pick and choose from different routers, template engines, IoT containers etc. But that increases the complexity and confusion as only some components work with some components. 

With ZF2 we had:
* a one, central point of contact,
* a desired/expected consistent architecture,
* handling of modularity, plugins, etc.
* some hard-coupling around MVC.

Now with recent efforts we're releasing the grip on the MVC and made it less coupled, BUT the **structure** seems to hold its ground and now compete directly with "Expressive". Expressive seems to be doing a catch-up game to ZF2: similar end-goals (easy to use architecture, extensibility, web-app framework) but different approaches and "missing features".

With current direction and velocity I can only see more confusion. There isn't now a single "architecture design" that is either recommended or preferred. I understand the benefits of staying neutral and respecting developers' choices but I can feel this becoming a major pain point esp. for less experienced developers who are not so opinionated nor knowledgeable about PSR7 vs event-driven MVC. 

Secondly, think about the version management - bumping major version (ZF2 to ZF3) is a perfect moment to break BC, introduce fresh new concepts and sweep legacy under the rug. ZF1 to ZF2 path was a good example of it. It was painful from migration perspective, but it enabled a quality jump like no other. I'm afraid we might be lose the benefit of this new-version moment, increase the overall complexity, bump up learning curve and muddle up the messaging even more.

My proposed solution is hard on the implementation side (all your points taken), is "opinionated", but simplifies and converges a lot in user-land; is better long-term.

module/Application/src/Module.php
samsonasik: I created PR https://github.com/weierophinney/ZendSkeletonApplication/pull/6 in case we will remove ModuleRouteListener usage.

samsonasik:
@weierophinney the wording and logo still uses 'ZF2', I will start provide patch with logo and send pull request to your pull request :dancer:

adamdyson:
@Thinkscape Well said and sadly I have to agree.

> @weierophinney There's two major concerns I see.
> 
> Even with extremely well constructed, immaculate PR play you still end up having to explain "Zend Framework" (the parent github.com/zendframework group and project marketing name). What is it then?
> 
> If you follow the "bag of loosely-coupled components" rhetoric, then you're allowed to give out several (competing) implementations of say, MVC. Thanks to interop efforts, it now goes even beyond github.com/zendframework and one can pick and choose from different routers, template engines, IoT containers etc. But that increases the complexity and confusion as only some components work with some components.
> 
> With ZF2 we had:
> 
> a one, central point of contact,
> a desired/expected consistent architecture,
> handling of modularity, plugins, etc.
> consistent education path,
> some hard-coupling around MVC.
> Now with recent efforts we're releasing the grip on the MVC and made it less coupled, BUT the structure seems to hold its ground and now compete directly with "Expressive". Expressive seems to be doing a catch-up game to ZF2: similar end-goals (easy to use architecture, extensibility, web-app framework) but different approaches and "missing features".
> 
> With current direction and velocity I can only see more confusion. There isn't now a single "architecture design" that is either recommended or preferred. I understand the benefits of staying neutral and respecting developers' choices but I can feel this becoming a major pain point esp. for less experienced developers who are not so opinionated nor knowledgeable about PSR7 vs event-driven MVC.
> 
> Secondly, think about the version management - bumping major version (ZF2 to ZF3) is a perfect moment to break BC, introduce fresh new concepts and sweep legacy under the rug. ZF1 to ZF2 path was a good example of it. It was painful from migration perspective, but it enabled a quality jump like no other. I'm afraid we might be lose the benefit of this new-version moment, increase the overall complexity, bump up learning curve and muddle up the messaging even more.
> 
> My proposed solution is hard on the implementation side (all your points taken), is "opinionated", but simplifies and converges a lot in user-land; is better long-term.

weierophinney:
@Thinkscape —

Sorry for the lateness of this response; I've been busy on the documentation milestone, and missed your comments until today, when I'm finally returning to this. Below are my thoughts on your latest comment. I *really* appreciate the feedback, even if I don't agree with it.

> If you follow the "bag of loosely-coupled components" rhetoric, then you're allowed to give out several (competing) implementations of say, MVC. Thanks to interop efforts, it now goes even beyond github.com/zendframework and one can pick and choose from different routers, template engines, IoT containers etc.

Yes and no. This has theoretically *always* been possible, due to the event-driven nature of `Zend\Mvc\Application`. Want a different routing solution? Register a custom "route" listener to do so. Want a different dispatcher? Register a custom "dispatch" listener (we already did with the `MiddlewareListener`!). etc. The efforts so far with the ZF3 initiatives have been to take this to its logical conclusion:

- Reduce dependencies/coupling. This makes using ZF components more palatable for many, who were reluctant because any single component was often pulling in up to a dozen others.
- Simplify wiring. We've been doing this by introducing `ConfigProvider` and `Module` classes to each component, providing the necessary configuration to wire them into IoC containers and, if required, the zend-mvc event manager.  zend-component-installer automates wiring these into zend-mvc- or Expressive-based applications.

The point is to allow developers the ability to pick and choose what they want, whether they are building Zend Framework applications *or not*. This includes Expressive.

> But that increases the complexity and confusion as only some components work with some components.

I'm not sure I agree with this statement at all. The work done to define the config providers and `Module` classes, as well as the component installer, makes certain that these play nicely with one another. Due to current efforts by the container-interop group, we'll likely have a solution around container configuration soon that will make this trivial *even for non-Zend application stacks*.

> With ZF2 we had:
> 
> - a one, central point of contact,
> - a desired/expected consistent architecture,
> - handling of modularity, plugins, etc.
> - consistent education path,
> - some hard-coupling around MVC.

Why are all of these considered "good"? And why do you assume they're all gone?

- I'm not sure what you mean by "one, central point of contact." If you mean a monolithic repository, my argument to that is [a storify on multi-repository benefits](https://storify.com/mwop/zf-components). The executive summary: a monolithic repository slows down and complicates release cycles, and doesn't scale from a maintentance standpoint. The move to many repositories has already proved itself; we've done multiple hundreds of releases since last fall, vs the dozen or so in the previous **two years**.

- While we can point to an "expected" or "consistent" architecture with ZF2, at the same time, I can argue it was not always "desired." Many developers disliked the default skeleton, as it was too complex for smaller tasks. Others followed some very strange practices when it came to directory layout. Some missed the single "library" or "src" directory of the ZF1 skeleton, and would recreate it in a ZF2 application. The point is: we can point the way, but we cannot enforce it, and anything we do to enforce it makes the structure inflexible.

- Modularity and plugins are still present. Literally the only difference with these currently is how modules are handled, as there are slight differences between zend-modulemanager and the proposed zend-config-manager (the latter handles literally only the configuration aggregation, arguably the most interesting and useful feature of zend-modulemanager). 

- Yes, we *do* lose a consistent education path to a degree, particularly if we have multiple skeletons (and we already do: one for zend-mvc, another for Expressive). However, this also gives us *flexibility*: we can have different skeletons *for different application purposes*. These each pose a potentially different educational approach. This could turn out to be either good or bad, but I think it's too early to claim either at this point.

- Hard-coupling around zend-mvc *is precisely what we needed to remove*. This alone was causing huge problems with keeping components updated and independent. It took us literally months to untangle that and make it possible for components to develop independently — and thus evolve. Tight coupling between components should be done only when strictly necessary. Tight coupling with the MVC makes it too hard to evolve our components, or allow them to be used easily outside the framework. I, for one, want to see *greater* adoption, not less.

> Expressive seems to be doing a catch-up game to ZF2: similar end-goals (easy to use architecture, extensibility, web-app framework)

Yes, but the approaches *are very different*. zend-mvc is an event-driven pipeline, using priorities, and requiring developers learn:

- IoC
- Our event system (which is a behemoth)
- How priority systems work (in order to correctly and effectively use the event system)
- Our module system (in order to learn how to do literally *anything* with the system, you need to build modules)
- Our HTTP abstraction layer (which differs from everybody else's, dis-allowing re-use between frameworks)

Expressive, on the other hand:

- Recommends and requires IoC, *but you can use it without*.
- Uses middleware pipelines for shaping application flow, *just like any other middleware framework*.
- Allows developers to structure the application according to application/organizational needs.
- Uses a standardized HTTP abstraction layer, allowing re-use of middleware between frameworks.

Catch-up game? In many ways, *zend-mvc is the one playing catch-up*.

*(I'll note that I still find ZF2 powerful, and a huge leap over ZF1. I also have take some time to analyze it critically — via the issues we've received, contact forms, stack overflow questions, in-person conversations, etc. — and I'm well aware of its _failings_, too. It's never fun tearing apart something you've poured your heart and soul into, but it's how we improve.*

> but different approaches and "missing features".

Most of those "missing features" are not missing at all, just delivered *by third-party framework-agnostic PSR-7 middleware libraries*. Which is exactly the point of having PSR-7 around in the first place — to enable such an ecosystem, so that no one framework or library ends up holding all the code.

> With current direction and velocity I can only see more confusion. There isn't now a single "architecture design" that is either recommended or preferred. I understand the benefits of staying neutral and respecting developers' choices but I can feel this becoming a major pain point esp. for less experienced developers who are not so opinionated nor knowledgeable about PSR7 vs event-driven MVC.

This is the place I agree with you. Having a choice in which application structure to choose *is* problematic. The choice was not made lightly, and there were several distinct considerations made:

- The ZF1 -&gt; ZF2 transition was a nightmare. Applications written to target ZF1 cannot be easily migrated to ZF2. End of story. I know many consultants, including Zend's own, who are *still working with clients* to migrate. Many chose *not* to migrate, and either are sticking with ZF1, fully knowing it will EOL eventually, or have chosen to jump ship to a different framework altogether.

  I, for one, did not want a repeat of this. So we needed to keep *some* version of zend-mvc around, *and make it as backwards compatible as possible*, to allow users to update.

- We've witnessed a huge change in the PHP ecosystem. Yes, Symfony has made huge in-roads in adoption, but an interesting, under-reported fact: a lot of that adoption has been through *other* frameworks such as Silex and Laravel, who build on top of it, and use a subset of the full-stack. Developers want something a) simpler, and/or b) more flexibly. PSR-7 middleware frameworks are hitting that sweet-spot, because they allow developers to compose their applications out of *any* libraries they want, and because they allow developers to define *any* workflow they want, simply and programmatically, instead of via an event abstraction. This is powerful.

> Secondly, think about the version management - bumping major version (ZF2 to ZF3) is a perfect moment to break BC, introduce fresh new concepts and sweep legacy under the rug. ZF1 to ZF2 path was a good example of it.

Please see my notes above. While it's a chance to increase quality and/or decrease complexity, it's also a chance to completely lose our existing user base, and likely ensure we never get adoption. What I'm trying to do here with this proposal is keep a semblance of the structure we've had during the ZF2 lifecycle, for our existing users and those developers who are fluent and comfortable with the ZF2 application workflow. For *new* projects, by *developers new to Zend Framework*, we will recommend Expressive, as a more modern, more flexible, and ultimately easier stack on which to build applications.

module/Application/src/Module.php
weierophinney: Interestingly, more than one pro services consultant has recently asked about `ModuleRouteListener` with recent zend-mvc releases, as clients were using it. I'll remove it, but I'm going to make a migration note that it's removed for new skeletons, and that users should be using fully qualified class names when referring to controllers and middleware.

weierophinney:
I've updated the issue description to add a TODO list (which is also in the branch).

weierophinney:
I've created [zend-skeleton-installer](https://github.com/zendframework/zend-skeleton-installer) for handling the "optional dependencies" installation requirements. It is a composer plugin that will prompt based on the contents of the `composer.json` extra.zend-skeleton-installer key, allowing users to choose the features they want to install in their project. Once installed, it uses zend-component-installer to update application configuration, and then removes itself from the project.

With that in place, I should be able to push the final updates (which will only be `composer.json` changes) required to complete this pull request and prepare it for testing and review.

weierophinney:
9c67fba adds the prompts into the `composer.json`; I then needed to update the zend-mvc dependency to alias it to 2.8.0, as zend-test and zend-mvc-console are pinned to `^2.7` versions; I'll get these updated before tagging zend-mvc 3.0 and having this patch merged.

At this point, the remaining tasks are:

- [ ] Update zend-mvc-console to pin to `^3.0.0-dev || ^3.0.0`
- [ ] Update zend-test to pin to `^2.7.1 || ^3.0`
- [ ] Add zend-development-mode as a development requirement
  - [ ] add `config/development.config.php.dist` as done in apigility
  - [ ] add `config/modules.config.php` as done in apigility
  - [ ] add verbiage post-install to execute `public/index.php development enable`
- [ ] Add verbiage before prompting configuration changes that users can safely choose to "Remember this option for other packages of the same type" (no packages are development *modules*).
- [ ] Document the changes to the skeleton either in zend-mvc or a tutorial (this likely can and will happen after merging this pull request)

weierophinney:
I'm going to defer the zf-development-mode additions to a later iteration, as the component needs to be updated to act as a console script (vs MVC console tooling) first.

I'm still mulling the prompt to tell users they can safely choose the "remember this option for other packages of the same type"; I'm thinking this would need to be a special target within the `OptionalPackagesInstaller` or other post-install-cmd listener in the zend-skeleton-installer. Regardless, it would be a change in that repository, not this one.

Documentation changes will need to take place in the tutorials; I've [created a wiki page detailing documentation needs](https://github.com/zendframework/maintainers/wiki/Tutorials-and-documentation-tasks-for-v3).

As such, I'd argue this is no longer a WIP, but can be tested and reviewed.

samsonasik:
I've tried, working as expected.

weierophinney:
Found more TODO items:

- [ ] Update the Vagrantfile to use ubuntu 16.04 and PHP 7.

Jodaille:
Hello,
I am not an expert, adding an example of object injection could be very useful.
I have tried to follow what I can understand with a really basic model, a factory in this fork:
https://github.com/Jodaille/ZendSkeletonApplication
Is it the right way to use/call a model in a controller ?

weierophinney:
@Jodaille — adding such examples is out-of-scope for the skeleton application, and belongs in the tutorials.

Regarding your example, it's spot on, though I'd suggest one change: implement `Zend\ServiceManager\Factory\FactoryInterface`, and modify your factory to implement `__invoke()` as it defines. It would then look something like the following:

```php
<?php
// IndexControllerFactory.php
namespace Application\Controller\Factory;

use Interop\Container\ContainerInterface;
use Zend\ServiceManager\Factory\FactoryInterface;
use Application\Controller\IndexController;

class IndexControllerFactory implements FactoryInterface
{
    public function __invoke(ContainerInterface $container, $name, array $options = null)
    {
        return new IndexController($container->get('MyApplication'));
    }
}
```

weierophinney:
Updated `Vagrantfile` created that:

- Uses the official `ubuntu/trusty64` as the base box (existing issues with VirtualBox and the `ubuntu/xenial64` box prevent its usage at this time)
- Updates to PHP 7.
- Installs, but does not run, composer (installed to `/usr/local/bin/composer`).
- Maps port 8080 instead of 8085 in order to be consistent with the Docker image.
- Uses the default vhost on the virtual machine, instead of creating a new one.
- Updates the default vhost to point its DocumentRoot at `/var/www/public`.
- Maps the current working directory to `/var/www`.

Adds a `Dockerfile` that:

- Builds on the php:7.0-apache image
- Installs the PHP zip extension (so that composer can run)
- Updates the default vhost on the virtual machine to point its DocumentRoot at `/var/www/public`.
- Installs composer to `/usr/local/bin/composer`.
- Makes the working directory and volume `/var/www`.
- Exposes port 80.

Adds a `docker-composer.yml` that:

- Uses the `Dockerfile` as its base.
- Builds against the working directory.
- Maps the working directory to the image's `/var/www`.
- Maps the host's port 8080 to the image's port 80.

weierophinney:
@zendframework/community-review-team Ready for review!

Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM php:7.0-apache
+
+RUN apt-get update \
+ && apt-get install -y git zlib1g-dev \
+ && docker-php-ext-install zip \
```
Ocramius: Why is the zip extension needed?

Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+ && a2enmod rewrite \
+ && sed -i 's!/var/www/html!/var/www/public!g' /etc/apache2/apache2.conf \
+ && mv /var/www/html /var/www/public \
+ && curl -sS https://getcomposer.org/installer \
```
Ocramius: Do you actually need composer in the container? From what it looks like here, installation will be run outside the container anyway, no?

README.md
```diff
@@ -1,118 +1,171 @@
...
+interfaces.
 
-### Installation using a tarball with a local Composer
+**Note:** The built-in CLI server is *for development only*.
 
-If you don't have composer installed globally then another way to create a new ZF2 project is to download the tarball and install it:
+## Using Vagrant
 
-1. Download the [tarball](https://github.com/zendframework/ZendSkeletonApplication/tarball/master), extract it and then install the dependencies with a locally installed Composer:
+This skeleton includes a `Vagrantfile` based on ubuntu 14.04, and using the
```
Ocramius: Can we just use 16.04? I must say that I don't see ubuntu boxes for 16.04 on vagrantbox.es though...

README.md
```diff
@@ -1,118 +1,171 @@
...
+$ vagrant ssh -c 'cd /var/www ; composer install'
+```
 
-If you don't have access to curl, then install Composer into your project as per the [documentation](https://getcomposer.org/doc/00-intro.md).
+While this will update them:
 
-Web server setup
-----------------
+```bash
+$ vagrant ssh -c 'cd /var/www ; composer update'
```
Ocramius: Should probably be `&&` (should `cd` fail, for whatever reason)

README.md
```diff
@@ -1,118 +1,171 @@
...
+## Using Docker
 
-    php -S 0.0.0.0:8080 -t public/ public/index.php
+This skeleton provides a `Dockerfile` for use with Docker. Build the image
+using:
 
-This will start the cli-server on port 8080, and bind it to all network
-interfaces.
+```bash
+$ docker build -t <name> .
```
Ocramius: `<name>` should probably be `<a-name-that-you-like>`

Dockerfile
weierophinney: For composer to run.

Dockerfile
weierophinney: You will if you want to create the project via `git clone` or downloading a tarball. This gives the option of not requiring a global composer.

README.md
```diff
@@ -1,118 +1,171 @@
...
+```
 
-**Note:** The built-in CLI server is *for development only*.
+Where `<name>` is a unique name for the image on your machine (this will usually
+be your project name). Once built, you can run it. To do so, you should map a
+port on your host machine to the image's port 80, and map the current directory
+to `/var/www` on the image:
+
+```bash
+$ docker run -d -p 8080:80 -v $(pwd):/var/www <name>
```
Ocramius: Since `$(pwd)` is used, we should specify that the command should be run from the project root directory

README.md
weierophinney: Nope, tried it. VirtualBox 5.0.18 has a known bug when running `ubuntu/xenial64` (I actually ran into it today when testing). We can update later once the issues are sorted.

README.md
```diff
@@ -1,118 +1,171 @@
...
+You can also run composer from the image. You will need to map your current
+directory to `/var/www` on the image when doing so:
 
-### Vagrant server
+```bash
+$ docker run -v $(pwd):/var/www <name> composer install
+```
 
-This project supports a basic [Vagrant](http://docs.vagrantup.com/v2/getting-started/index.html) configuration with an inline shell provisioner to run the Skeleton Application in a [VirtualBox](https://www.virtualbox.org/wiki/Downloads).
+Remembering to map the volume and port when running vanilla Docker can become
```
Ocramius: Since `docker-compose` is available: do we need to even teach `docker run` here?

TODO.md
```diff
@@ -0,0 +1,43 @@
+# TODO
+
+This is a TODO list for the feature/zend-mvc-v3-minimal branch.
+
+## Installer
+
+- Prompt for caching
```
Ocramius: These are solved TODOs

README.md
weierophinney: I already do - I mention "map the current directory" explicitly...

composer.json
```diff
@@ -1,14 +1,117 @@
...
     "name": "zendframework/skeleton-application",
-    "description": "Skeleton Application for ZF2",
+    "description": "Skeleton Application for Zend Framework",
     "license": "BSD-3-Clause",
     "keywords": [
         "framework",
         "zf2"
     ],
     "homepage": "http://framework.zend.com/",
+    "minimum-stability": "dev",
```
Ocramius: We should get rid of this, no?

README.md
weierophinney: Is docker-compose installed by default when installing docker now? If so, I can remove those instructions; if not, it's probably good to have vanilla docker instructions just in case.

module/Application/config/module.config.php
```diff
@@ -1,104 +1,60 @@
...
-    ),
-    'view_manager' => array(
+                    ],
+                ],
+            ],
+        ],
+    ],
+    'controllers' => [
+        'factories' => [
+            Controller\IndexController::class => InvokableFactory::class,
```
Ocramius: Symbol to be imported?

module/Application/src/Module.php
```diff
@@ -0,0 +1,18 @@
...
+ * @link      http://github.com/zendframework/ZendSkeletonApplication for the canonical source repository
+ * @copyright Copyright (c) 2005-2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace Application;
+
+class Module
+{
+    const VERSION = '3.0.0dev';
```
Ocramius: Argh, I wish we were on 7.0 only :-P

module/Application/view/application/index/index.phtml
```diff
@@ -1,43 +1,43 @@
 <div class="jumbotron">
-    <h1><?php echo sprintf($this->translate('Welcome to %sZend Framework 2%s'), '<span class="zf-green">', '</span>') ?></h1>
-    <p><?php echo sprintf($this->translate('Congratulations! You have successfully installed the %sZF2 Skeleton Application%s. You are currently running Zend Framework version %s. This skeleton can serve as a simple starting point for you to begin building your application on ZF2.'), '<a href="https://github.com/zendframework/ZendSkeletonApplication" target="_blank">', '</a>', \Zend\Version\Version::VERSION) ?></p>
-    <p><a class="btn btn-success btn-lg" href="https://github.com/zendframework/zf2" target="_blank"><?php echo $this->translate('Fork Zend Framework 2 on GitHub') ?> &raquo;</a></p>
+    <h1><?php echo sprintf('Welcome to %sZend Framework%s', '<span class="zf-green">', '</span>') ?></h1>
```
Ocramius: These view scripts were really horrible to read. Can we simplify them a bit? All this `sprintf()` feels really unnecessary

composer.json
weierophinney: Need it for now if you want to use zend-test; without this set, that component will not install (and it needs to be pinned to dev-develop until zend-mvc v3 is released, which is when we can do a zend-test v3 release).

public/index.php
```diff
@@ -14,8 +17,12 @@
     unset($path);
 }
 
-// Setup autoloading
-require 'init_autoloader.php';
+// Composer autoloading
+include 'vendor/autoload.php';
```
Ocramius: We should use an absolute path here, to avoid reliance on the include path

module/Application/config/module.config.php
weierophinney: Already is, line 12.

public/index.php
```diff
@@ -14,8 +17,12 @@
...
+
+if (! class_exists(Application::class)) {
+    throw new RuntimeException('Unable to load ZF2. Run `php composer.phar install`.');
+}
 
 // Run the application!
-Zend\Mvc\Application::init(require 'config/application.config.php')->run();
+Application::init(require 'config/application.config.php')->run();
```
Ocramius: Same here: I suggest moving to an absolute path

module/Application/src/Module.php
weierophinney: :smile:

On the flip side, at least we're moving to only two supported PHP versions for now!

Dockerfile
Ocramius: 👍

README.md
Ocramius: Alright. Sadpanda about stability of VirtualBox

README.md
Ocramius: Ah, yep, I misread that

README.md
Ocramius: I don't know about linux environments, but on windows and mac (docker beta), `docker-compose` is part of the bundle

composer.json
Ocramius: So this is a pre-merge TODO, or is it a chicken-egg problem?

module/Application/config/module.config.php
Ocramius: Yes, but you are using an imported namespace, not the class. That's what I was referring to. This is mostly for consistency

README.md
weierophinney: Alrighty; removing the "Docker" section momentarily.

composer.json
weierophinney: chicken-egg problem. We can keep it as is, and it will install stable packages when detected. I setup the constraints to prefer stable packages as well (`^3.0 || ^3.0.0-dev@dev`), so leaving it will work for now. Once zend-mvc and zend-test are tagged, we can remove the line. (And that will happen before we prepare this repository for release.)

module/Application/config/module.config.php
weierophinney: Imported namespace?

The file is in the `Application` namespace, and `Controller` is a subnamespace of it...

module/Application/view/application/index/index.phtml
weierophinney: Oh, good point! Those were there due to translations, but are largely unnecessary now. Working on it now.

module/Application/config/module.config.php
Ocramius: Yup, maybe it's just my OCD kicking in, but I'd expect `use Application\Controller\IndexController`. Preferences, at this point: your pick

Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+  | php -- --install-dir=/usr/local/bin --filename=composer
+
+WORKDIR /var/www
+
+VOLUME /var/www
+
+EXPOSE 80
```
Maks3w: Parent image already expose this port. If this is for documentation porpuses we could just comment write as a comment.

Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM php:7.0-apache
+
+RUN apt-get update \
```
Maks3w: Is this Dockerfile intended for use in production?

Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+ && docker-php-ext-install zip \
+ && a2enmod rewrite \
+ && sed -i 's!/var/www/html!/var/www/public!g' /etc/apache2/apache2.conf \
+ && mv /var/www/html /var/www/public \
+ && curl -sS https://getcomposer.org/installer \
+  | php -- --install-dir=/usr/local/bin --filename=composer
+
+WORKDIR /var/www
+
+VOLUME /var/www
```
Maks3w: This is already done by docker-compose. I wish to remove this definition from here as does not promove self contained containers.

Vagrantfile
```diff
@@ -5,35 +5,22 @@ VAGRANTFILE_API_VERSION = '2'
...
+echo "    vagrant ssh -c 'cd /var/www ; composer install'"
+echo "** [ZF] Visit http://localhost:8080 in your browser for to view the application **"
 SCRIPT
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
-  config.vm.box = 'bento/ubuntu-14.04'
-  config.vm.network "forwarded_port", guest: 80, host: 8085
-  config.vm.hostname = "skeleton-zf.local"
-  config.vm.synced_folder '.', '/var/www/zf'
+  config.vm.box = 'ubuntu/trusty64'
```
Maks3w: gbarbieru/xenial include the patches needed for work with current vagrant releases.

Vagrantfile
```diff
@@ -5,35 +5,22 @@ VAGRANTFILE_API_VERSION = '2'
 
 @script = <<SCRIPT
 DOCUMENT_ROOT_ZEND="/var/www/zf/public"
```
Maks3w: This env var is now unused

module/Application/config/module.config.php
Maks3w: I understand that requires more mental effort to understand the left side is a relative namespace while the right side is an alias.

In theory the developer should rename Application namespace to something like Vendor\Component and extend this config with more and more self crafted classes. This implies lot of repetition of Vendor\Component many times.

The most common case for factories could be
 `Vendor\Component\SubComponent\Class:class => Vendor\Component\SubComponent\ClassFactory:class`
which could easily ends as a very large line.

I think we can continue as is and change this in the feature if needed. Did we receive any comment for this since we put the namespace declaration in this file?

Vagrantfile
Maks3w: Append this if you want for change the default workdir. this allow to remove all those `cd /var/www ;` from vagrant ssh commands

```
    if ! grep "cd /vagrant" /home/vagrant/.profile > /dev/null; then
      echo "cd /vagrant" >> /home/vagrant/.profile
    fi
```

public/index.php
```diff
@@ -14,8 +17,12 @@
     unset($path);
 }
 
-// Setup autoloading
-require 'init_autoloader.php';
+// Composer autoloading
+include __DIR__ . '/../vendor/autoload.php';
+
+if (! class_exists(Application::class)) {
+    throw new RuntimeException('Unable to load ZF2. Run `php composer.phar install`.');
```
Maks3w: This is the first time `composer.phar` term is used. I think we should write something like

> Type vagrant ssh `composer install` if you are using vagrant. Type docker exec composer install if you are using docker. Type composer install or php composer.phar if you are working in local`

README.md
```diff
@@ -1,118 +1,141 @@
...
+```bash
+$ docker-compose build
+```
 
-### Vagrant server
+Once built, you can run it:
 
-This project supports a basic [Vagrant](http://docs.vagrantup.com/v2/getting-started/index.html) configuration with an inline shell provisioner to run the Skeleton Application in a [VirtualBox](https://www.virtualbox.org/wiki/Downloads).
+```bash
+$ docker-compose up -d
```
Maks3w: Both docker commands can be simplified as: `docker-compose up -d --build`

README.md
```diff
@@ -1,118 +1,141 @@
...
+```
 
-1. Run vagrant up command
+At this point, you can visit http://localhost:8080 to see the site running.
```
Maks3w: Probably this need an special mention for those systems which need a docker VM.

Dockerfile
weierophinney: Most likely you would end up modifying it for production to include your application specific needs. The primary intent is to allow developers to start with a consistent development environment.

Dockerfile
weierophinney: When I started, I was providing both a vanilla `Dockerfile` *and* docker-compose configuration separately. I'll remove it now, as we've decided to recommend docker-compose specifically.

Vagrantfile
weierophinney: I'd rather pin to official repositories. We can update this later once virtualbox/vagrant are patched and the new releases in wide distribution.

Dockerfile
Maks3w: I wonder if we could split the command for make it more simple and add more things to docker cache even if it add more layers

weierophinney: This is doing exactly three things:

- Ensuring we have the extensions necessary to run composer.
- Ensuring the web server is configured correctly (enables mod_rewrite and changes the document root).
- Installing composer.

I'm not sure we really need to simplify it further…

Maks3w:
This is ok for me :+1:

weierophinney:
## Status update

First, I've updated to stable versions of:

- zend-mvc (3.0)
- zend-test (3.0)
- zend-mvc-console (1.1.10, which marks zend-mvc v2 as a conflict)
- zend-mvc-i18n (1.0)
- zend-mvc-plugins (1.0; see below)

This puts the pull request in a good position to be tagged once merged. However…

I've created a [zend-mvc-plugins](https://github.com/zendframework/zend-mvc-plugins) metapackage, to aggregate all official plugins, reducing the number of prompts by three. However, on experimentation, it appears that packages installed via a metapackage *do not trigger plugins*; instead, only the *metapackage* triggers the plugin.

The solution will be to allow the `module` and `component` keys to be arrays, which will allow zend-mvc-plugins to define the list of components for zend-component-installer to install. *Until that is done, this PR cannot be merged.*

weierophinney:
## Status update

- I've released [zend-component-installer 0.2.0](https://github.com/zendframework/zend-component-installer/releases/tags/0.2.0), which adds support for arrays of components, modules, and config-providers when defining packages.

- I've released [zend-skeleton-installer 0.1.1](https://github.com/zendframework/zend-skeleton-installer/releases/tags/0.1.1), which fixes a bug in the `Uninstaller` whereby the `composer.lock was not updated to reflect the removal of the skeleton installer package.

- I've released [zend-mvc-plugins 1.0.1](https://github.com/zendframework/zend-mvc-plugins/releases/tags/1.0.1), which updates the package to define an array of components for the component installer to inject into configuration.

At this time, this patch is finally ready to merge!

weierophinney:
Pushed to develop for release with 3.0.0.

PR #330, opened 2016-04-22
default http response code is 500, so fatal errors return correct res…

prolic:
…ponse code

### DIFF
--- a/public/index.php
+++ b/public/index.php
@@ -14,6 +14,9 @@ if (php_sapi_name() === 'cli-server') {
     unset($path);
 }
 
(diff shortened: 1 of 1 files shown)


samsonasik:
👍

Maks3w:
This only happens when you have XDebug enabled.

prolic:
@Maks3w nope.

Try putting this in your index.php

```
<?php

throw new Exception('foo');
```

Disable xdebug and hit the browser, you'll see the exception and a http status code 200.

Maks3w:
Declined by consistency with https://github.com/zfcampus/zf-apigility-skeleton/pull/83

PR #337, opened 2016-06-09
Add skeleton unit tests

alextech:
To try lowering barrier of entry to figuring out how to test MVC application, as witnessed in up-voted questions on ZF documentation discussion threads, and to try establish better mindset for having tests as part of the process from the beginning, thought I would add skeleton tests complementing current skeleton functionality.

Open to discussion on

- configuration layout and appropriate key overrides, if any
- adequacy of suggested assertions


froschdesign:
@alextech 
Is this really needed in the skeleton application? We already have a chapter in the tutorials: "[Unit Testing a zend-mvc application](https://zendframework.github.io/tutorials/unit-testing/)"

alextech:
This is a step that is always missing in the workflow. Unlike other configurations and sources that get added based on what optional components user needs, this one has to be done in all cases after each installation. And it has to be done specific same way each time to reflect the state of the skeleton, regardless of what user plans to do with it next.  Unlike other components, with tutorials or not, testing sources have to mirror app sources exactly, instead of going their own way. Therefore, I would argue that the user should not have to figure out that whole tutorial first just to mirror what zend-skeleton-installer did as initial step: zend-test tutorial should be a compliment to understand what the installer did and be able to fill in tests as more controllers, actions, or other components get added.

Argument that testing is optional, that it can be left out and done at user's discretion in separate tutorial, does not fit today's ZF's and most PHP projects' culture, in my opinion. With "no test - no (PR/bugreport)" rules and promotions, skeleton should reflect the same and not make it an afterthought.  I believe tutorials and quick start skeletons should have them integrated as a single flow of information, not "lets push a bunch of code out to solve the problem"  followed by "ok now that everything is working lets do it all over again with tests" - it will not be taken seriously.

rbroen:
Please do merge, it is in line with: test early, test often

froschdesign:
@alextech and @rbroen 
Don't get me wrong. I am on your side – nothing without tests.
But in the past there were many discussions on this: What is a "skeleton" and what should be included (or not).

@weierophinney 
[Can we combine this with the installer?](https://github.com/zendframework/ZendSkeletonApplication/blob/develop/composer.json#L88)

weierophinney:
I like this idea, quite a lot, actually.

Three comments:

- Use `test` not `tests` for the testing directory (singular form is the convention used for both source file and test file directories on all ZF components, and most packages I've reviewed the last few years).
- Rename `phpunit.xml` to `phpunit.xml.dist`, and move it to the top-level directory. Again, it's the leading convention, and allows developers to toggle env flags in a local file without altering the file checked into source control. Having it in the application root simplfies invocation (no need to know the path to the test configuration file). When you do, add `phpunit.xml` to the `.gitignore` rules.
- Add an `autoload-dev` rule for the new `ApplicationTest` namespace.

We may also want to add travis configuration at some point so we can verify that changes we make to the skeleton are reflected in the tests over time, but that can happen later.

alextech:
Thank you for the detailed corrections: learned good details.  Very happy you find it useful.

Regarding your comment "changes we make to the skeleton are reflected in tests over time" and commits I saw coming today, I was thinking to what extent some of the view templates should be covered.  Specifically 404.phtml with its view logic.  I did a very simple layout related one as a sample because checking all scope variables could make it way too long for a skeleton going beyond it's scope like froschdesign original concern but at same time don't want to do this half way. 
If I do not come up with a short elegant way to cover it before merge, may do another PR.

phpunit.xml.dist
```diff
@@ -0,0 +1,7 @@
+<phpunit colors="true">
```
samsonasik: missing `bootstrap="vendor/autoload.php"` ?

alextech: With bare skeletal functionality, that would be redundant since phpunit/phpunit/phpunit shipped in the vendor directory to execute the tests does this. It populates as far as I understand, it populates PHPUNIT_COMPOSER_INSTALL which is then used by phpunit framework to autoload classes.

Since in ZF3 we do not use zend autoloader, bootstrapping framework there is also not needed.

However, if want to customize test runner more, maybe with some utility functions or reusable mocks to help bring up application as a common place for setUp() then it can be created as an empty stub ahead of time open to those extensions. Would that be beyond the scope of this or we stick to what works bare minimum?

samsonasik: oh, just realize  that, it's  ok then ;)

README.md
```diff
@@ -29,6 +29,27 @@ interfaces.
...
+as a development dependency with
+
+```bash
+$ composer require --dev zendframework/zend-test
+```
+
+Once installed, you can run the tests using:
+
+```bash
+$ vendor/phpunit/phpunit/phpunit
```
alextech: How many times used it, did not realize this

.gitignore
```diff
@@ -9,3 +9,4 @@ nbproject
 .settings
 vendor
 composer.phar
+phpunit.xml.dist
```
samsonasik: `phpunit.xml` instead of `phpunit.xml.dist`

alextech: i had that fixed but in excitement forgot to pull when went to work computer and overwrote it :(
Good catch.

weierophinney:
@alextech Many thanks for this feature!

PR #338, opened 2016-06-09
Added additional instanceof checks for Error types

weierophinney:
Since zend-mvc 3.0 will catch any Throwable when under PHP 7, we need to test for either `Exception` or `Error` in the error handling view scripts.

See https://github.com/zendframework/zend-mvc/issues/157#issuecomment-224987189 for an example of why this is needed.

### DIFF
--- a/module/Application/view/error/404.phtml
+++ b/module/Application/view/error/404.phtml
@@ -48,7 +48,7 @@ if (isset($this->controller_class)
 
 <?php if (isset($this->display_exceptions) && $this->display_exceptions): ?>
 
-<?php if(isset($this->exception) && $this->exception instanceof Exception): ?>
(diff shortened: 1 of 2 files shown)


module/Application/view/error/404.phtml
```diff
@@ -48,7 +48,7 @@ if (isset($this->controller_class)
 
 <?php if (isset($this->display_exceptions) && $this->display_exceptions): ?>
 
-<?php if(isset($this->exception) && $this->exception instanceof Exception): ?>
+<?php if(isset($this->exception) && ($this->exception instanceof Exception || $this->exception instanceof Error)): ?>
```
weierophinney: Just a note: we cannot typehint against just `Throwable`, as the skeleton also supports PHP 5.6. As such, testing for either `Exception` or `Error` will satisfy any `Throwable` type from PHP 7.

Issue #334, opened 2016-06-04
Missing Template Map Generator


RalfEggert:
I am missing the template map generator after the installation of the new Skeleton Application which is ready for ZF3. I do not want to require https://github.com/zendframework/zendframework because that has all the dependencies listed.

Ocramius:
Does https://github.com/Ocramius/OcraCachedViewResolver work in this context? I think it's a better alternative than generating template maps manually...

Ocramius:
Note: didn't upgrade the module to 3.x component, so I do not know if it works

RalfEggert:
It cannot be installed:

```
$ composer require ocramius/ocra-cached-view-resolver
You are running composer with xdebug enabled. This has a major impact on runtime performance. See https://getcomposer.org/xdebug
Using version ^3.0 for ocramius/ocra-cached-view-resolver
./composer.json has been updated
Loading composer repositories with package information
Updating dependencies (including require-dev)
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - ocramius/ocra-cached-view-resolver 3.1.x-dev requires zendframework/zend-mvc ~2.3 -> satisfiable by zendframework/zend-mvc[2.3.0, 2.3.1, 2.3.2, 2.3.3, 2.3.4, 2.3.5, 2.3.6, 2.3.7, 2.3.8, 2.3.9, 2.4.0, 2.4.0rc1, 2.4.0rc2, 2.4.0rc3, 2.4.0rc4, 2.4.0rc5, 2.4.0rc6, 2.4.0rc7, 2.4.1, 2.4.10, 2.4.2, 2.4.3, 2.4.4, 2.4.5, 2.4.6, 2.4.7, 2.4.8, 2.4.9, 2.5.0, 2.5.1, 2.5.2, 2.5.3, 2.6.0, 2.6.1, 2.6.2, 2.6.3, 2.7.0, 2.7.1, 2.7.2, 2.7.3, 2.7.4, 2.7.5, 2.7.6, 2.7.7, 2.7.8] but these conflict with your requirements or minimum-stability.
[...]

PR #340, opened 2016-06-11
Added Template Map Generator

RalfEggert:
Added template map geneartor to solve #334

### DIFF
--- a/bin/templatemap_generator.php
+++ b/bin/templatemap_generator.php
@@ -4,7 +4,7 @@
  * Zend Framework (http://framework.zend.com/)
  *
  * @link      http://github.com/zendframework/zf2 for the canonical source repository
- * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
+ * @copyright Copyright (c) 2005-2016 Zend Technologies USA Inc. (http://www.zend.com)
  * @license   http://framework.zend.com/license/new-bsd New BSD License
  */
 
--- a/bin/templatemap_generator.php
+++ b/bin/templatemap_generator.php
@@ -1,0 +1,256 @@
+#!/usr/bin/env php
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+use Zend\Console;
+use Zend\Loader\StandardAutoloader;
+
+/**
+ * Generate template maps.
+ *
+ * Usage:
+ * --help|-h                    Get usage message
+ * --library|-l [ <string> ]    Library to parse; if none provided, assumes
+ *                              current directory
+ * --view|-v [ <string> ]       View path to parse; if none provided, assumes
+ *                              view as template directory
+ * --extensions|-e [ <string> ] List of accepted file extensions (regex alternation
+ *                              without parenthesis); default: *
+ * --output|-o [ <string> ]     Where to write map file; if not provided,
+ *                              assumes "template_map.php" in library directory
+ * --append|-a                  Append to map file if it exists
+ * --overwrite|-w               Whether or not to overwrite existing map file
+ */
+
+// Setup/verify autoloading
+if (file_exists(__DIR__ . '/../vendor/autoload.php')) {
+    // Local install
+    require __DIR__ . '/../vendor/autoload.php';
+} elseif (file_exists(getcwd() . '/vendor/autoload.php')) {
+    // Root project is current working directory
+    require getcwd() . '/vendor/autoload.php';
+} elseif (file_exists(__DIR__ . '/../../../autoload.php')) {
+    // Relative to composer install
+    require __DIR__ . '/../../../autoload.php';
+} else {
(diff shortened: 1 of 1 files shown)


bin/templatemap_generator.php
```diff
@@ -0,0 +1,256 @@
+#!/usr/bin/env php
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
```
samsonasik: 2016

RalfEggert: Done

bin/templatemap_generator.php
```diff
@@ -0,0 +1,256 @@
+#!/usr/bin/env php
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
```
samsonasik: https://github.com/zendframework/ZendSkeletonApplication

Ocramius: Can we please leave these things out of the skeleton? Seems excessive to
me, at least...

vrkansagara:
@Ocramius  :+1:

bin/templatemap_generator.php
RalfEggert: But where should the template map generator be placed then? It doesn't make sense to add the zendframework/zendframework package just to get the template map generator. Create a new package just for the generator? Seems excessive to me... ;-)

RalfEggert: Ok, thought about it a little more. How about adding the template map generator to https://github.com/zendframework/zend-view?

samsonasik: I think @Ocramius comment is about placing license header, not about where template map generator

weierophinney: @RalfEggert I like that idea; that puts it in the same repo that would consume what it generates.

If you do this, don't forget to add the `bin` entry for it in the `composer.json` (per the [Vendor Binaries documentation](https://getcomposer.org/doc/articles/vendor-binaries.md)). Usage would be like any other installed binary: `./vendor/bin/templatemap_generator.php ...`.

Ping me when you have a PR! :smile:

bin/templatemap_generator.php
```diff
@@ -0,0 +1,256 @@
...
+ * @copyright Copyright (c) 2005-2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+use Zend\Console;
+use Zend\Loader\StandardAutoloader;
```
weierophinney: This last import statement can be removed.

weierophinney: Which brings me to an interesting dilemma... The script makes use of `Zend\Console\Getopt`, which would mean that zend-view would need to add zend-console as a dependency, which I doubt we want to do.

I think we could likely simplify the script to have the signature:

```bash
$ templatemap_generator.php <directory>
```

and have it spit what it generates to STDOUT; developers can then pipe that to a file. As an example, to overwrite:

```bash
$ templatemap_generator.php module/Album/view > config/autoload/views.global.php
```

To append:

```bash
$ templatemap_generator.php module/Album/view >> config/autoload/views.global.php
```

Alternately, we could require that they pass the full set of files via globbing, which would accommodate the current `--extensions` argument:

```bash
$ templatemap_generator.php module/Album/view/**/*.phtml >> module/Album/config/template_map.php
```

If the script assumes that the map should be relative to the `PWD`, then we could suggest this workflow:

```bash
$ cd module/Album/view
$ ../../../vendor/bin/templatemap_generator.php ./**/*.phtml > ../config/template_map.php
```

Calling the script with no arguments or with `--help` or `-h` would provide the usage message.

The above essentially accommodates all current options, and would allow eliminating the dependency on zend-console; we'd only have to parse `$argv` and grab the first argument.

Thoughts?

bin/templatemap_generator.php
RalfEggert: @weierophinney 

Here you go: https://github.com/zendframework/zend-view/pull/65

I hope everything is ok, since I am sitting in a train home and don't have access to my main note book. I only have a windows machine here right now.

I guess this PR can be closed if you accept the one at zend-view

bin/templatemap_generator.php
RalfEggert: Maybe add this discussion to the https://github.com/zendframework/zend-view/pull/65?

bin/templatemap_generator.php
Ocramius: > I think @Ocramius comment is about placing license header, not about where template map generator

No, I actually am against having the template map generator in the skeleton.

First of all, it could live in a separate module, like all other MVC-ish modules in here.
Second, I already built something that completely removes the need for a template map, and is compatible with v3 - https://github.com/Ocramius/OcraCachedViewResolver

This, in my opinion, is micro-optimization and complication.

RalfEggert: @Ocramius 

Well, the template map generator will be moved to zend-view, anyway. 

I like your OcraCachedViewResolver but since it requires PHP 7 it is no alternative as long as ZF3 components have a minimum requirement of PHP 5.6.

RalfEggert:
Fixed by https://github.com/zendframework/zend-view/pull/67

PR #341, opened 2016-06-15
Fix route type

autowp:
'Zend\Mvc\Router\Http\Literal' -> 'Literal'


kokspflanze:
this is already fixed in develop
https://github.com/zendframework/ZendSkeletonApplication/blob/develop/module/Application/config/module.config.php#L18

Issue #342, opened 2016-06-15
Update error views


webimpress:
Guys,
what do you think about updating error views (`404.phtml` and `index.phtml`)?
I don't like this old style...

For example closing tag is not in the same column what opening tag:
https://github.com/zendframework/ZendSkeletonApplication/blob/develop/module/Application/view/error/404.phtml#L35-L44

missing some indents for if/while bodies:
https://github.com/zendframework/ZendSkeletonApplication/blob/develop/module/Application/view/error/index.phtml#L27-L60

and generally the whole style of these files isn't so nice.

What do you think? Do you prefer to use `if (...): ... endif;` or better with braces `{ ... }` ?

alextech:
bad indentation for templates is pretty annoying, especially if it produces badly indented output to browser. Fixing that would be nice.
using `if (..): ... endif;` is a commonly accepted convention for HTML mix since its easier to trace in IDE than bracket matching.

weierophinney:
Fixed recently with #344.

PR #344, opened 2016-06-15
Error views updated

webimpress:
Related to https://github.com/zendframework/ZendSkeletonApplication/issues/342

* fixed indents and spacing
* converted all `<?php echo` statements to `<?=`
* removed redundant classes in error views
* `$ex` for exception variable instead of `$e`
* used constant instead of strings
* added `escapeHtml` on error message and trace

Any suggestions would be appreciated.

### DIFF
--- a/module/Application/view/error/404.phtml
+++ b/module/Application/view/error/404.phtml
@@ -1,7 +1,7 @@
 <h1>A 404 error occurred</h1>
 <h2><?= $this->message ?></h2>
 
-<?php if (!empty($this->reason)): ?>
+<?php if (!empty($this->reason)) : ?>
     <?php
     switch ($this->reason) {
         case \Zend\Mvc\Application::ERROR_CONTROLLER_CANNOT_DISPATCH:
@@ -27,7 +27,7 @@
     <p><?= $reasonMessage ?></p>
 <?php endif ?>
 
-<?php if (isset($this->controller) && $this->controller): ?>
+<?php if (isset($this->controller) && $this->controller) : ?>
     <dl>
         <dt>Controller:</dt>
         <dd>
@@ -44,8 +44,8 @@
     </dl>
 <?php endif ?>
 
-<?php if (isset($this->display_exceptions) && $this->display_exceptions): ?>
-    <?php if (isset($this->exception) && ($this->exception instanceof \Exception || $this->exception instanceof \Error)): ?>
+<?php if (isset($this->display_exceptions) && $this->display_exceptions) : ?>
+    <?php if (isset($this->exception) && ($this->exception instanceof \Exception || $this->exception instanceof \Error)) : ?>
         <hr/>
 
         <h2>Additional information:</h2>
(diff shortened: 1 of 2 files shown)


alextech:
nice! alignment made such a great difference. So is classes clean up

module/Application/view/error/404.phtml
```diff
@@ -1,113 +1,105 @@
...
-        break;
-    default:
-        $reasonMessage = 'We cannot determine at this time why a 404 was generated.';
-        break;
-}
-?>
-
-<p><?= $reasonMessage ?></p>
-
+<?php if (!empty($this->reason)): ?>
```
samsonasik: should use `{` instead of `:` for consistencies? also, Psr2

alextech:
Thank a lot @webimpress for the cleanup!
I am now migrating an application to ZF3 and noticing that error 500 pages, ones that trigger 500.phtml template, always skip _if(isset($this->exception) && $this->exception instanceof Exception):_ condition, displaying _No Exception available_.   I see it is fixed in 404.phtml at https://github.com/zendframework/ZendSkeletonApplication/pull/338 so since you are already looking at cleaning error pages, thought would let you do it one time, instead of opening another PR myself.

alextech:
A separate request regarding class names. 
So far in all my experience error pages horribly clash with style sheets of any custom templates, eg. ones bought from themeforest. Do you think can add some class names (eg. class="error") to different error field string, or at least to parent HTML container, so that applying applying theme is a little less painful?

module/Application/view/error/404.phtml
webimpress: ... for consistencies with ...? I can't see in any other views loops or conditional statements (where php is mixed with html). And it was like that before `if...endif`.
I've ran PHPCS with PSR-2 standards on views, and if I add spaces after parenthesis (before `:`) everything looks fine. I'm fine with change everything to `{` but then we have decide that is new standard from now and maybe add some PSR-2 checks/auto fixes also on views?

BTW. Is zend using PSR-2 at all? I'm not sure, because of (for example): https://github.com/zendframework/zend-mvc/blob/master/.php_cs#L13

webimpress:
@alextech sorry, how do you mean? It is there already:
https://github.com/zendframework/ZendSkeletonApplication/pull/344/files#diff-162067fcf8cbfae3a0fc7ea971e15954R48

Regarding class names, I'm not sure if it's needed. This is just application skeleton. Guys, any thoughts?

alextech:
Hm, I saw 500.phtml had it in my setup, but just realized it is not in upstream new skeleton anymore. Assuming it was decided to not have 500 anymore, never mind.

If you do not think errors should not be semantically distinguished in the UI layer from regular content and surrounding layout, its easy enough to customize myself I guess if others do not agree.

weierophinney:
As noted when we were working on the Apigility skeleton, there's one change I'll make when merging with regards to indentation.

The PR indents HTML at the same level as the PHP conditionals. However, this leads to some very strangely formatted markup when done:

```html
<h1>An error occurred</h1>
<h2>Some error message</h2>
        <hr/>
        <h2>Additional information:</h2>
        <h3>Some\Exception\Type</h3>
        ...
                    <hr />
                    <h2>Previous exceptions:</h2>
                    <ul class="list-unstyled">
                            <li>...</li>
```

etc. The change I'll be making is to indent the HTML to match the actual DOM hierarchy it's in. This will mean that it's slightly less readable as source code, but more readable as markup. You will be able to easily match levels of PHP indentation, however, in part due to the restructuring to use inline tags whenever possible.

Also, per a conversation with @webimpress on IRC, I'll update the conditionals to follow what we did in Apigility, changing structures like `if (isset($this->someProp) && $this->someProp)` to `if (! empty($this->someProp))`.

PR #345, opened 2016-06-21
Provide development mode

weierophinney:
zf-development-mode has been a popular feature of Apigility, and a popular standalone module. This patch adds v3 of that package, which has been re-engineered to work as a standalone script (vs as a console-based MVC controller). In particular:

- I've added `config/development.config.php.dist`, to allow specifying bootstrap-level configuration to use when development mode is enabled (including additional modules, changes in configuration caching, etc.)
- I've added `config/autoload/development.local.php.dist`, to allow specifying application-level configuration to use when development mode is enabled; by default, this includes enabling error display in zend-mvc views.
- I've split module listing out to `config/modules.config.php`, as we did in the Apigility skeleton. This provides a single file in which to view them, and simplifies auto-generation of that list.
- I've added three composer scripts to simplify working with development mode:
  - `composer development-enable`
  - `composer development-disable`
  - `composer development-status`

### Notes

This PR originally took the route of creating a script within the skeleton, `bin/development-mode`. That approach was scrapped in favor of producing an updated version of zfcampus/zf-development-mode that had no external dependencies.


bin/development-mode
```diff
@@ -0,0 +1,235 @@
...
+/**
+ * Disable development mode
+ *
+ * @return int
+ */
+function disable()
+{
+    if (! file_exists('config/development.config.php')) {
+        // nothing to do
+        echo "Development mode was already disabled.", PHP_EOL;
```
webimpress: why `"` not just `'` ?

bin/development-mode
```diff
@@ -0,0 +1,235 @@
...
+    if (file_exists('config/development.config.php')) {
+        // nothing to do
+        echo "Already in development mode!", PHP_EOL;
+        return 0;
+    }
+
+    if (! file_exists('config/development.config.php.dist')) {
+        fwrite(
+            STDERR,
+            "MISSING \"config/development.config.php.dist\". Could not switch to development mode!" . PHP_EOL
```
webimpress: why not:
`'MISSING "config/development.config.php.dist". Could not switch to development mode!'` ?

bin/development-mode
```diff
@@ -0,0 +1,235 @@
...
+
+if (! in_array($argv[1], ['disable', 'enable'], true)) {
+    fwrite(STDERR, 'Unrecognized argument.' . PHP_EOL . PHP_EOL);
+    fwrite(STDERR, $help . PHP_EOL);
+    exit(1);
+}
+
+$action = $argv[1];
+$return = $action();
+exit((int) $return);
```
webimpress: do we need this `(int)` type casting ?

weierophinney:
Ping @akrabat - thoughts? The other direction I could take is to do a v3 release of zend-development-mode that just ships this script, instead of acting as a module, and require that in the ZF skeleton.

bin/development-mode
weierophinney: @webimpress Maybe, maybe not. `exit()` expects an integer, however, and I'd rather code defensively; we may enable strict mode in the future, and this will ensure that works.

bin/development-mode
weierophinney: Because I copied this verbatim from zend-development-mode. :smile:

public/index.php
```diff
@@ -29,5 +30,11 @@
     );
 }
 
+// Retrieve configuration
+$appConfig = require __DIR__ . '/../config/application.config.php';
+if (file_exists(__DIR__ . '/config/development.config.php')) {
+    $appConfig = ArrayUtils::merge($appConfig, require __DIR__ . '/../config/development.config.php');
```
alextech: file_exists and require are checking for different files

weierophinney: Thanks, @alextech - fixing now!

webimpress:
The code in `development-mode` is quite long.... Wouldn't be nicer to write a class and create new instance with `$argv`?
```
class DevelopmentMode {
    public function __construct(array $argv)
    {
        ...
    }
}

new DevelopmentMode($argv);
```
 
I don't like this procedural style...

weierophinney:
> I don't like this procedural style...

Matter of preference. I don't care either way; this was relatively easy, and didn't require mixing a class and a script in a single file, or pushing a class into the skeleton source tree. If we were to do move this into zend-development-mode, I'd likely switch to a class, however.

bin/development-mode
webimpress: No point now, you call only one of two methods and both return `int`... so you cast int to int, always. Even more, I'm not sure (for example) if `null` response should be casted to 0, so if you want to have defensive code maybe better to have check if response `!is_int` ... ?

bin/development-mode
```diff
@@ -0,0 +1,235 @@
...
+{
+    static $dir;
+
+    if ($dir) {
+        return $dir;
+    }
+
+    $config = getApplicationConfig();
+    if (isset($config['module_listener_options']['cache_dir'])
+        && ! empty($config['module_listener_options']['cache_dir'])
```
webimpress: `!empty` is enough

bin/development-mode
```diff
@@ -0,0 +1,235 @@
...
+{
+    static $key;
+
+    if ($key) {
+        return $key;
+    }
+
+    $config = getApplicationConfig();
+    if (isset($config['module_listener_options']['config_cache_key'])
+        && ! empty($config['module_listener_options']['config_cache_key'])
```
webimpress: as above - `!empty` is enough, `isset` and `!empty` is redundant

bin/development-mode
```diff
@@ -0,0 +1,235 @@
...
+    return $key;
+}
+
+/**
+ * Return the application configuration.
+ *
+ * Memoizes the discovered configuration so subsequent calls can re-use the
+ * value.
+ *
+ * Exits with status code 2 if unable to find the configuration.
```
webimpress: I can see `exit(1)` below, not `2`

bin/development-mode
```diff
@@ -0,0 +1,235 @@
...
+}
+
+/**
+ * Return the configured configuration cache directory, if any.
+ *
+ * @return null|string
+ */
+function getConfigCacheDir()
+{
+    static $dir;
```
webimpress: no point, the method is used only once in the whole script

weierophinney:
Alternate version available in https://github.com/zfcampus/zf-development-mode/pull/19 (pinging @webimpress and @akrabat)

bin/development-mode
weierophinney: `! empty()` will raise an error if the key does not exist, while `isset()` will not (and will return `false` if empty); I've gone with just the `isset()` check in the zf-development-mode PR.

webimpress: sorry, `!empty()` definitely will not raise any errors even if the key does not exist
https://3v4l.org/3HtYl

> Determine whether a variable is considered to be empty. A variable is considered empty if it does not exist or if its value equals FALSE. empty() does not generate a warning if the variable does not exist.
http://php.net/empty

Still I think `!empty()` is better choice.

weierophinney:
I've updated this pull request to use zf-development-mode v3, and added several composer script aliases to simplify working with it.

README.md
```diff
@@ -29,6 +29,23 @@ interfaces.
...
 **Note:** The built-in CLI server is *for development only*.
 
+## Development mode
+
+The skeleton ships with [zf-development-mode](https://github.com/zfcampus/zf-development-mode)
+by default, and provides three aliases for consuming the script it ships with:
+
+```bash
+$ composer development-enable  # enable development mode
+$ composer development-disable # enable development mode
```
akrabat: s/enable/disable

akrabat:
Was afk all week. +1 from me, though note typo in README.

PR #346, opened 2016-06-22
Provide development mode - class

webimpress:
@weierophinney please have a look on my improvements - development mode as class.
There are still few things I don't like...

Why I'd prefer this solution?
* everything is encapsulated inside an object
* removed duplications - only one response method returns error and success response depends on response code
* removed  `static` variables
* it could be easier and cleaner parameterized
* code looks more friendly - no procedural code

I don't like:
* `response` method could be better
* maybe it's not clear that calling `response` method stop further execution
* calling on create object (maybe better static method `DevelopmentMode::exec(...)`?)
* maybe we need to pass only `$argv` instead of both: `$argc`, `$argv`

Maybe all of it is matter of preferences. I'd like to fine the most friendly and modern solution. Finally this is gonna be used with ZF3! I think we can improve it a bit 😃 

Thoughts?

### DIFF
--- a/bin/development-mode
+++ b/bin/development-mode
@@ -1,0 +1,235 @@
+#!/usr/bin/env php
+<?php
+/**
+ * @link      http://github.com/zendframework/ZendSkeletonApplication for the canonical source repository
+ * @copyright Copyright (c) 2005-2016 Zend Technologies USA Inc. (http://www.zend.com)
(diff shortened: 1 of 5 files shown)


webimpress:
Closed because of changes in https://github.com/zendframework/ZendSkeletonApplication/pull/345

PR #352, opened 2016-06-30
module version updated - no longer dev !

webimpress:
The version wasn't updated on release zf3 to 3.0.0, so everyone who start with new ZendSkeletonApplication still see `zf v3.0.0dev` :(

### DIFF
--- a/module/Application/src/Module.php
+++ b/module/Application/src/Module.php
@@ -9,7 +9,7 @@ namespace Application;
 
 class Module
 {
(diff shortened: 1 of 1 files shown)


webimpress:
I think maybe better will be to merge other PR and bump version to 3.0.1 than releasing this hotfix ...

/cc @Ocramius @weierophinney @GeeH

GeeH:
Let's get this sorted automatically as in #356 rather than go down the route of merging these manually.

PR #360, opened 2016-07-08
Let's disable development mode with development-disable

Forage:


Forage:
This is already going to be fixed by an earlier PR #348

PR #361, opened 2016-07-08
Alter DocumentRoot in 000-default.conf

Forage:
There's no `/var/www/html` text to be replaced in the default `/etc/apache2/apache2.conf` file.
This change will alter the same file as the Vagrant config.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ RUN apt-get update \
  && apt-get install -y git zlib1g-dev \
  && docker-php-ext-install zip \
  && a2enmod rewrite \
- && sed -i 's!/var/www/html!/var/www/public!g' /etc/apache2/apache2.conf \
+ && sed -i 's!/var/www/html!/var/www/public!g' /etc/apache2/sites-available/000-default.conf \
  && mv /var/www/html /var/www/public \
  && curl -sS https://getcomposer.org/installer \
   | php -- --install-dir=/usr/local/bin --filename=composer


weierophinney:
Have you experienced an actual issue here?

The `php:7.0-apache` image has an `apache2.conf` file that differs from the one normally installed by Debian/Ubuntu, and which contains a `DocumentRoot` entry for the default vhost; the configuration we're using here in the `Dockerfile` updates that entry specifically.

As a test, I just did the following:

- Cloned the repository to a new directory
- Ran `composer install`
- Ran `composer development-enable`
- Ran `docker-compose build`
- Ran `docker-compose up`
- Visited localhost:8080

All worked as expected, and I was able to see that apache was accepting and responding to my requests.

jasonsemko:
Can we please get this merged in?

I used:
```
composer create-project zendframework/skeleton-application test-zend-apache
```

To install, get 404s unless I do this fix.

webimpress:
I think the issue is valid.
@weierophinney please check your local docker images:
```bash
$ docker images
```
and try update it to the latest version:
```bash
$ docker pull php:7.0-apache
```

For more information please check my comment https://github.com/zfcampus/zf-apigility-skeleton/pull/150#issuecomment-262752400.

michaellenahan:
I needed this patch today to get the docker container to work.

Without the patch:
```
docker-compose down
docker-compose build
docker-compose up
```
http://localhost:8080/
```
Not Found
The requested URL / was not found on this server.
Apache/2.4.10 (Debian) Server at localhost Port 8080
```
With the patch:
```
docker-compose down
docker-compose build
docker-compose up
```
http://localhost:8080/
... The Skeleton Application is served.

Xerkus:
Confirmed

PR #363, opened 2016-07-14
Update Dockerfile

sillasleal:
Line 7( && sed -i 's!/var/www/html!/var/www/public!g' /etc/apache2/apache2.conf \) does not work properly with Apache2 on Debian


sillasleal:
The fix is only for those who use the Docker

PR #365, opened 2016-07-16
Update composer create-project command

mahkali:
Missing a space


webimpress:
It is fine also without space. Composer is using `symfony/console` component, and here you are what are the allowed arguments formats:
http://symfony.com/doc/current/components/console/console_arguments.html

mahkali:
Ok great. Thanks for the link.

PR #366, opened 2016-07-16
Fixed missing interface

mahkali:
Added `implements ProviderInterface` to the Application Module

### DIFF
--- a/module/Application/src/Module.php
+++ b/module/Application/src/Module.php
@@ -7,7 +7,7 @@
 
 namespace Application;
 
-class Module
+class Module implements ConfigProviderInterface
 {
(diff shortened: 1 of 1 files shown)


weierophinney:
We specifically chose not to do this, as in upcoming releases of various components and zend-modulemanager, we will be moving the interfaces to the component they target. As such, to keep forwards compatibility, we are not implementing the interfaces.

This works, as the various feature listeners in zend-modulemanager look for either explicit implementations, or duck-typed implementations.

You are free to update it in your new project, and to implement the interfaces in your own modules, but we'll be keeping it out of the skeleton for now.

Issue #356, opened 2016-07-01
Update version number as part of build process


GeeH:
We need to to this as part of release but I have no idea how.

Suggestions.

Ocramius:
Somebody backport and maintain `ocramius/package-versions` for PHP 5.x plz :trollface:

GeeH:
Banning @Ocramius for blatant advertisment.

GeeH:
On a serious note, would you have a problem with this repository having a dependency on `ocramius/package-versions` @weierophinney?

Ocramius:
`ocramius/package-versions` is (purposely) 7.0.x only, but since it's MIT, anyone can just grab it and backport the functionality. As it stands, it's just about removing few type hints and strict type declarations

GeeH:
It's easier if we just up the requirements of this skeleton application to 7.0.x.

stijnhau:
That might be true but i think de dependcy on PHP-7 for the default app that 90% of people will start with is just to soon.

samsonasik:
@Ocramius I forked your `ocramius/package-versions` and try to make it compatible to ^5.6 in my forked repo ( https://github.com/samsonasik/PackageVersions/pull/1 ) , would you mind to review ?

samsonasik:
there is a samsonasik/package-versions now! a backport of ocramius/package-versions that support php 5.6 ;)

samsonasik:
I've created PR #371 for it to use samsonasik/package-versions as backport of ocramius/package-versions that support php 5.6 ;)

weierophinney:
And... despite having this and related issues open today... I forgot to bump the version when I released 3.0.3 a short bit ago, belying the need for this to be in place.

I'm still unsure how to proceed. The version constant approach does have use cases in terms of us helping understand where to start with a user, but typically the output of `composer show` will be far more relevant. Having that info in the browser _could_ be useful, particularly for users who are uncomfortable with the command line and/or composer, but means an extra package that serves no purpose once the template is replaced.

In many ways, I think I'd rather just drop the version constant altogether. Thoughts?

webimpress:
@weierophinney 
> In many ways, I think I'd rather just drop the version constant altogether. Thoughts?

[...]

PR #371, opened 2016-07-21
use samsonasik/package-versions to get version

samsonasik:
`samsonasik/package-versions` is a backported version of `ocramius/package-versions` that support php 5.6. /cc @Ocramius

### DIFF
--- a/module/Application/view/application/index/index.phtml
+++ b/module/Application/view/application/index/index.phtml
@@ -4,7 +4,7 @@
     <p>
         Congratulations! You have successfully installed the
         <a href="https://github.com/zendframework/ZendSkeletonApplication" target="_blank">ZF Skeleton Application</a>.
-        You are currently running Zend Framework version <?= \Application\Module::VERSION ?>.
+        You are currently running Zend Framework 3 with zend-mvc version <?= \PackageVersions\Versions::getVersion('zendframework/zend-mvc') ?>.
         This skeleton can serve as a simple starting point for you to begin
         building your application on ZF.
     </p>
--- a/module/Application/view/application/index/index.phtml
+++ b/module/Application/view/application/index/index.phtml
@@ -9,6 +9,14 @@
         building your application on ZF.
     </p>
 
+    <p>Here are the list of ZF components installed</p>
+    <?php $zendframeworkComponents = array_slice(scandir('./vendor/zendframework'), 2); ?>
+    <ul>
+        <?php foreach ($zendframeworkComponents as $component) { ?>
+            <li><?php echo 'zendframework/' . $component . ':' . \PackageVersions\Versions::getVersion('zendframework/'.$component) ?></li>
+        <?php } ?>
+    </ul>
+
     <p><a class="btn btn-success btn-lg" href="https://github.com/zendframework/zendframework" target="_blank">Fork Zend Framework on GitHub &raquo;</a></p>
 </div>
 
--- a/composer.json
+++ b/composer.json
@@ -13,6 +13,7 @@
     "prefer-stable": true,
     "require": {
         "php": "^5.6 || ^7.0",
+        "samsonasik/package-versions": "^1.1",
         "zendframework/zend-component-installer": "^1.0 || ^0.3 || ^1.0.0-dev@dev",
         "zendframework/zend-skeleton-installer": "^1.0 || ^0.1.3 || ^1.0.0-dev@dev",
         "zendframework/zend-mvc": "^3.0.1",
--- a/composer.lock
+++ b/composer.lock
@@ -4,8 +4,8 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "68b34965034959dd0cc938584e52fc9d",
-    "content-hash": "bbc4cf51d583edd647a090e9419d33ef",
+    "hash": "344c1c1bab6b8b19213b0e9021f965ff",
+    "content-hash": "11fd59139a5bd1e68f64a704113911de",
     "packages": [
         {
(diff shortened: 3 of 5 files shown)


webimpress:
@samsonasik it is not the same as before.
Now we have `zend-mvc` version, before it was (should be) `ZendSkeletonApplication` version.

I don't know if the constant version in the library is important (useful) at all.
It could help investigate issues (in ZendSkeletonApplication) if we know what version of ZendSkeletonApplication was used by developer. For me, we should have fixed "Zend Framework 3" as we had for ZF2.

samsonasik:
unfortunatelly, the packageversions can't detect its version of project currently used, as if i use zendframework/zend-skeleton-application (not part of vendor), it will return like 999999.dev, probably @Ocramius can give some input about it...

Warm regards,

Abdul Malik Ikhsan

Pada 21 Jul 2016, pukul 21.59, webimpress <notifications@github.com> menulis:

Ocramius:
That simply means "dev-master@dev"

samsonasik:
@Ocramius any suggestion of what it should be then?

Warm regards,

Abdul Malik Ikhsan

Pada 22 Jul 2016, pukul 02.12, Marco Pivetta <notifications@github.com> menulis:

> That simply means "dev-master@dev"
>

Ocramius:
`999999.dev@SHA1YADDAYADDA` is correct

samsonasik:
`\PackageVersions\Versions::getVersion('zendframework/skeleton-application')` return `9999999-dev@89c54ddbed3c6a17bc20fbc74218d7d60efaa8f4`, what `$packageName` should be passed then?

Ocramius:
Probably `'zendframework/zend-mvc'` ;-)

samsonasik:
I did that :P

module/Application/view/application/index/index.phtml
```diff
@@ -4,7 +4,7 @@
     <p>
         Congratulations! You have successfully installed the
         <a href="https://github.com/zendframework/ZendSkeletonApplication" target="_blank">ZF Skeleton Application</a>.
-        You are currently running Zend Framework version <?= \Application\Module::VERSION ?>.
+        You are currently running Zend Framework 3 with zend-mvc version <?= \PackageVersions\Versions::getShortVersion('zendframework/zend-mvc') ?>.
```
samsonasik: `\PackageVersions\Versions::getVersion('zendframework/zend-mvc')` is ok to get full version with hash included

weierophinney:
This is a hard one.

With ZF1/2 (until 2.5), `Zend\Version\Version::VERSION` was available, and denoted the framework version. The move to the constant in the skeleton was prompted by the fact that the only singular version we can look at is the skeleton itself.

However, as @samsonasik has noticed, `PackageVersions` works on the dependencies, but not the application — and even if it did, it would change immediately, as the version then becomes that of the application, not the skeleton!

The point we need to discuss is: is the information worthwhile?

I'd argue that, to an extent, it is: it tells us where the application *began*, and most times we're getting that, we're working with developers new to composer and/or ZF; if they report this particular version, it helps us understand that we're likely working with newcomers, and whether or not they started with the latest skeleton version.

If that's the case, the only real way we can version is either through a constant in the application somewhere, or one defined in the `composer.json` (in the `extra` tree).

Thoughts?

webimpress:
In my opinion if we tried limit number of dependencies in the skeleton we should't include any package version here. Also I can't see big advantage of having the constant version number, even if we are working with newcomers we don't know what steps they did before:
- `composer install` or `composer update`?
- minimal installation or with optional packages?
- any custom configuration/dependencies?

Installed packages we could get from `composer show` and maybe we should add it to documentation. If somebody will have issue with base skeleton it means our release is broken, what is quite unlikely (hopefully). I'd remove constant and the dependency (of `PackageVersions`) and, in case of issue, ask for the list of installed packages (`composer show`) or/and steps to reproduce the issue.

samsonasik:
@weierophinney I'm ok with show list of zf components installed, if approved.
@webimpress it is common that user seeing framework version number when they installed the skeleton, so version number show is needed imo.

samsonasik:
@weierophinney I added list of zf components installed.
list-of-zf-components-installed

samsonasik:
@weierophinney merge-able now ?

adamlundrigan:
I'm not a fan of adding it directly into the skeleton as non-dev since it's not aimed at the application the user wants to build...it's a convenience for the developer and for those of us who help them when they run into trouble.  Could the package be added to require-dev and still fulfill #356?

As for the UI, maybe this is better pushed over to ZendDeveloperTools?  It could have a panel which lists all the installed versions.  

> \<mwop>it tells us where the application began, and most times we're getting that, we're working with developers new to composer and/or ZF; if they report this particular version, it helps us understand that we're likely working with newcomers, and whether or not they started with the latest skeleton version.

I do agree this would be useful.  The developer isn't likely to continually backport changes from newer versions of the skeleton into their existing apps, so the files provided by the skeleton end up being frozen in time while the Composer-provided dependencies do.  Knowing what version of the skeleton they started with, then, is a useful piece of information.  I'd argue that the starting versions of the skeleton's dependencies aren't, though (eg: is it useful to know that servicemanager was at 2.6.0 when they installed the skeleton but it's at 2.7.7 now?). 

Could we instead use composer scripts to cache this information somewhere on install and/or update?  The DX around that could be improved by ZendDeveloperTools and/or CLI tooling to display the cached version information.

Xerkus:
for debugging purposes we can ask output from `composer show`, otherwise i tend to agree with @adamlundrigan  that extra dependencies like this are not really needed in a skeleton application

Xerkus:
Since no one else decided to move forward with this feature and overall consensus is towards dropping version constant altogether, I am going to close this PR without merging.

PR #375, opened 2016-08-11
Vagrant Ubuntu 16.04 LTS

webimpress:
Vagrant configuration updated to Ubuntu 16.04 LTS (ubuntu/xenial64).

Main user is now `ubuntu` (previously was `vagrant`).
Added also name for the vagrant box: "ZF Application - Ubuntu 16.04".

Log file `ubuntu-xenial-16.04-cloudimg-console.log` is created in the main directory. Not sure if we can disable it somehow. Any ideas?

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ control. (If you want to make the modifications permanent, edit the
 
 ## Using Vagrant
 
-This skeleton includes a `Vagrantfile` based on ubuntu 14.04, and using the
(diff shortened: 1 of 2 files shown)


weierophinney:
> Log file ubuntu-xenial-16.04-cloudimg-console.log is created in the main directory. Not sure if we can disable it somehow. Any ideas?

I started noticing that as well when testing the Apigility skeleton. From the research I've seen, it's specific to VirtualBox, and likely to the new 5.1 versions. I'll add a gitignore rule for that file.

PR #377, opened 2016-08-15
QA Tools

webimpress:
As we have it in apigility skeleton I've added CS rules and composer scripts: `cs-check`, `cs-fix`, `test`.
Base test for `Application\Controller\IndexController` and phpunit dist configuration was already there.
Updated notes in README.md about script `composer serve`.

Also added section "QA Tools' into README.md.

Maybe we should add another note in README.md - `zend-test` is one of the optional packages of the installer - hm... or maybe we should change "Running Unit Tests" section, or remove this part at all.

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -1,0 +1,32 @@
+<?xml version="1.0"?>
+<ruleset name="Zend Framework coding standard">
+    <description>Zend Framework coding standard</description>
+
+    <!-- display progress -->
+    <arg value="p"/>
+    <arg name="colors"/>
+    <arg name="extensions" value="php,dist,phtml"/>
+
+    <!-- inherit rules from: -->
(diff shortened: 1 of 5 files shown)


phpcs.xml
```diff
@@ -0,0 +1,32 @@
...
+        <properties>
+            <property name="ignoreBlankLines" value="false"/>
+        </properties>
+    </rule>
+    <rule ref="PSR1.Files.SideEffects">
+        <exclude-pattern>public/index.php</exclude-pattern>
+    </rule>
+
+    <!-- Paths to check -->
+    <file>config</file>
```
weierophinney: We should add the `.dist` files to this; I'll do that on merge.

weierophinney: Never mind; I see you have them configured in the extensions.

PR #380, opened 2016-08-15
IIS URLrewrite configuration

webimpress:
Added `web.config` file to support IIS URLrewrite (like the .htaccess)
Reference: <https://framework.zend.com/manual/1.12/en/project-structure.rewrite.html>

As added in apigility skeleton in https://github.com/zfcampus/zf-apigility-skeleton/commit/c5f7e1957d7d99a462e66605319bc21645f12cd3

### DIFF
--- a/public/web.config
+++ b/public/web.config
@@ -1,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
+    <system.webServer>
+        <rewrite>
+            <rules>
+                <rule name="Imported Rule 1" stopProcessing="true">
+                    <match url="^.*$" />
+                    <conditions logicalGrouping="MatchAny">
+                        <add input="{REQUEST_FILENAME}"
+                             matchType="IsFile" pattern=""
+                             ignoreCase="false" />
+                        <add input="{REQUEST_FILENAME}"
+                             matchType="IsDirectory"
+                             pattern=""
+                             ignoreCase="false" />
+                    </conditions>
+                    <action type="None" />
+                </rule>
+                <rule name="Imported Rule 2" stopProcessing="true">
+                    <match url="^.*$" />
+                    <action type="Rewrite" url="index.php" />
+                </rule>
+            </rules>
+        </rewrite>
+    </system.webServer>
+</configuration>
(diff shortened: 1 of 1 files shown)


public/web.config
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
+    <system.webServer>
+        <rewrite>
+            <rules>
+                <rule name="Imported Rule 1" stopProcessing="true">
```
alextech: Better descriptive names.

webimpress: @alextech good point :) changed! Thanks

alextech:
Good addition. Microsoft IIS could use some attention since they have been very helpful to PHP community themselves.

I have been curious about that reference doc, what does the first rule accomplish that second cannot?  When I switch to IIS a few months ago, I get by fine with just

```
        <rewrite>
            <rules>
                <rule name="Forward request to non-existentrequest to ZF application" stopProcessing="true">
                    <match url="^.*$" />
                    <conditions logicalGrouping="MatchAll">
                        <add input="{REQUEST_FILENAME}" matchType="IsFile" ignoreCase="false" negate="true" />
                    </conditions>
                    <action type="Rewrite" url="index.php" />
                </rule>
            </rules>
        </rewrite>
```

Can you comment on whether my version is missing out some corner case out the documented version covers or is documented version unnecessary bloated?

Also wondering on explicitly specifying which I have in mine.
```
        <defaultDocument>
            <files>
                <clear />
                <add value="index.php" />
            </files>
        </defaultDocument>
```

Would it be safer to add or redundant?

weierophinney:
@alextech The submitted one is more thorough. In particular, it is checking if the matched URI references *either* a file *or* a directory on the filesystem; in the latter case, IIS will check to see if a default/index document exists and serve it. In the rule you submitted, directories that match will always be redirected to the application `index.php`.

Your rules could pretty easily be updated to include this condition, from what I can see. That said, I find the rules submitted easier to follow, in part because I don't have to spend effort seeing what is *excluded* from a given rule, only what the rule *matches*.

alextech:
@weierophinney your explanations are always so amazingly written! Thank you, will switch to the ones submitted here.

Issue #368, opened 2016-07-19
Disable config caching by default


olegkrivtsov:
I tried to install the latest version of Zend Skeleton Application. I installed session support, standard plugins and form support. Then I opened the skeleton in web browser (Chrome). Then I added the `aboutAction()` action to `IndexController` and a Literal route for that action and a view template. So far so good.

Then I tried to open the About page in my browser, I've received error "The URL couldn't be matched by routing". Confusing. After many trials and errors, I've realised that disabling the config caching in `application.config.php` fixes the problem.

So, I suggest that the config caching be disabled by default (currently it is enabled).

samsonasik:
you can run:
```
$ composer development-enable
```

olegkrivtsov:
I think that caching should be disabled BY DEFAULT. When you download Zend Skeleton App, you definitely plan to develop your website from it, so config caching enabled by default may be a problem. You don't identify this problem in your documentation, so novice users will be confused.

samsonasik:
it is documented: http://zendframework.github.io/tutorials/getting-started/skeleton-application/#development-mode

GeeH:
Ping @weierophinney 

I agree with this. I would like to see development mode enabled by default so that no caching happens, then a "deploying to production" document explaining what you need to do to make an app production ready. 

Possible solution is to just disable the caching in the `application.config.php` with the shipped application. A better solution would be to run  `composer development-enable` as part of the create project process.

vrkansagara:
Agree with gray.

adamlundrigan:
Adding this to the `scripts` section of the `composer.json` should be sufficient:

```
"post-create-project-cmd": "@development-enable",
```

PR #381, opened 2016-08-16
Enable development mode after install is completed

adamlundrigan:
Implements #368


webimpress:
@adamlundrigan I think we should also update README.md

adamlundrigan:
Added a short blurb.  How does that look to you?

weierophinney:
@adamlundrigan That is AWESOME! I always forget that the `scripts` section can reference Composer events as well!

Thanks!

PR #390, opened 2016-09-30
Update Dockerfile to fix DocumentRoot config file

texai:

Issue #389, opened 2016-09-28
Dockerfile should also modify the document root in the default vhost


charlieman:
The Dockerfile changes apache's root directory to `/var/www/public` in apache2.conf but doesn't do it in the default virtual host configuration located in `/etc/apache2/sites-available/000-default.conf`. Changing the [sed line](https://github.com/zendframework/ZendSkeletonApplication/blob/a0049bb/Dockerfile#L7) to this should do it:

```
&& sed -i 's!/var/www/html!/var/www/public!g' /etc/apache2/apache2.conf \
  /etc/apache2/sites-available/000-default.conf \
```

adamlundrigan:
#390 fixes this but stops updating `apache2.conf`.  Is it necessary to set the root in both files?

charlieman:
I think in the default file is enough.

Mezzle:
Yes, the `000-default.conf` is enough. `/var/www/html` is not mentioned in the main conf

PR #391, opened 2016-10-08
	Fixed and modernised docker / docker-compose setup

rawkode:
- Switched to docker-compose syntax version: '2'
- Updating docker-compose to provide a tmpfs cache directory that won't sync / pollute local disk

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,11 @@
 FROM php:7.0-apache
 
 RUN apt-get update \
- && apt-get install -y git zlib1g-dev \
- && docker-php-ext-install zip \
- && a2enmod rewrite \
- && sed -i 's!/var/www/html!/var/www/public!g' /etc/apache2/apache2.conf \
- && mv /var/www/html /var/www/public \
- && curl -sS https://getcomposer.org/installer \
-  | php -- --install-dir=/usr/local/bin --filename=composer
+  && apt-get install -y git zlib1g-dev \
+  && docker-php-ext-install zip \
+  && a2enmod rewrite \
+  && sed -i 's!/var/www/html!/var/www/public!g' /etc/apache2/sites-enabled/000-default.conf \
+  && curl -sS https://getcomposer.org/installer \
+    | php -- --install-dir=/usr/local/bin --filename=composer
 
 WORKDIR /var/www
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,12 +1,11 @@
...
- && docker-php-ext-install zip \
- && a2enmod rewrite \
- && sed -i 's!/var/www/html!/var/www/public!g' /etc/apache2/apache2.conf \
- && mv /var/www/html /var/www/public \
- && curl -sS https://getcomposer.org/installer \
-  | php -- --install-dir=/usr/local/bin --filename=composer
+  && apt-get install -y git zlib1g-dev \
+  && docker-php-ext-install zip \
+  && a2enmod rewrite \
+  && sed -i 's!/var/www/html!/var/www/public!g' /etc/apache2/sites-enabled/000-default.conf \
```
webimpress: Please have a look on apigility version of this file:
https://github.com/zfcampus/zf-apigility-skeleton/blob/master/Dockerfile
We have there also:
` && mv /var/www/html /var/www/public \`

Isn't it required here?

rawkode: `docker-compose` mounts your local directory to `/var/www`, removing whatever is there in the base `Dockerfile`; we simply need to do the `sed` command to update the `VirtualHost` and everything is fine, as the `public` directory exists within the mount.

adamlundrigan:
#389 and #390 propose similar changes.  I vote this PR be accepted and the others closed.

rawkode:
@adamlundrigan  Anything I can do to help massage this through?

bitwombat:
The README.md also need to correct:

docker-compose up -d --build
to
docker-compose up -d

bitwombat:
It should be noted that this change requires docker-compose 1.6+.  Ubuntu 16.04 LTS has 1.5.2 currently.

Not sure what minimum versions of docker stuff the team wants to require.  Is docker-compose 2.0 backwards compatible?

The fixes to the sed are actually an error, and should probably merged right away, as in  #390

adamlundrigan:
@rawkode not sure who is the maintainer of this repository currently, but @weierophinney originally added the Docker stuff so I will tag him.

rawkode:
Shall I just close this? @weierophinney @adamlundrigan

PR #398, opened 2016-12-15
Add bower package manager and update application layout

elialejandro:
I suggest use bower and install all dependencies with this manager, the first update is with bootstrap


alextech:
Already denied in discussion https://github.com/zendframework/ZendSkeletonApplication/pull/339 and https://github.com/zendframework/ZendSkeletonApplication/pull/241

PR #407, opened 2017-02-03
Update Dockerfile

alec892:
The apache2.conf file hasn't the line "/var/www/html". The DocumentRoot of 000-default.conf file must to "/var/www/public"


webimpress:
Duplicate #390

PR #410, opened 2017-03-03
Fixed default DocumentRoot location

sergei-solomonov:
Default virtual host file `/etc/apache2/sites-available/000-default.conf`. Now `docker-compose up -d` works


webimpress:
Duplicate #390

sergei-solomonov:
Doh, I should check it before... :)

PR #413, opened 2017-03-14
Update composer serve instruction in README to account for default composer timeout

Xerkus:
`composer run --timeout 0 serve` will disable default 5 minutes timeout


ezimuel:
@Xerkus thanks!

PR #420, opened 2017-10-04
Update Module.php

rbroen:
Missing this in a fresh project will drive you absolutely crazy.
Please add this to the ZendSkeletonApplication.

### DIFF
--- a/module/Application/src/Module.php
+++ b/module/Application/src/Module.php
@@ -15,4 +15,15 @@ class Module
     {
         return include __DIR__ . '/../config/module.config.php';
     }
+    
+    public function getAutoloaderConfig()
+    {
+        return [
+            'Zend\Loader\StandardAutoloader' => [
+                'namespaces' => [
+                    __NAMESPACE__ => __DIR__ . '/src/',
+                ],
+            ],
+        ];
+    }
 }


alextech:
No, this is something composer does now. In composer.json you specify

    "autoload": {
         "psr-4": {
            "Application\\": "module/Application/src/",
            "Investment\\": "module/Investment/src/"
       }
    }

Followed by `composer dump-autoload`. Zend's autoloader bootstrapping nightmare should never return.

froschdesign:
And for the application module it is already there:

https://github.com/zendframework/ZendSkeletonApplication/blob/e233481d0496225b2dd5b994337d00f5f6188133/composer.json#L21-L25

rbroen:
The problem is that when users add a new module to their project (probably by copying the Application module), even after all changes are made in that module and the module is listed in modules.config.php, it will not work.

Putting ZF configuration in the composer is adding "magic" functionality. New users will not get it, things need to be more transparent. (Or the "magic" functionality should work for all modules in the module folder, but transparency would have my preference).

froschdesign:
> Putting ZF configuration…

This is no ZF configuration, it's only autoloading. See: https://getcomposer.org/doc/01-basic-usage.md#autoloading
No magic here, because Composer is an integral part of external libraries and internal modules / own application.

> New users will not get it, things need to be more transparent. 

We will consider this in the documentation and the quick start guide. Thanks for the hint!
(In the current documentation you will find an explanation in the chapter "[Zend Expressive - Modular applications](https://docs.zendframework.com/zend-expressive/features/modular-applications/#module-structure)".)

If you use "[zend-expressive-tooling](https://github.com/zendframework/zend-expressive-tooling)", then the scripts will do the job for you!

PR #428, opened 2018-03-03
add use_zend_loader = false in config/application.config.php

samsonasik:
to use composer autoloader instead of zend-loader


weierophinney:
Thanks, @samsonasik! Merged to develop for release with 3.1.0

PR #444, opened 2019-06-14
fix(module/Application/test) IndexControllerTest - setyp method have …

KOVCHENKO:
setUp method in module/Application/test/Controller/IndexControllerTest.php has been fixed

### DIFF
--- a/module/Application/test/Controller/IndexControllerTest.php
+++ b/module/Application/test/Controller/IndexControllerTest.php
@@ -13,7 +13,7 @@ use Zend\Test\PHPUnit\Controller\AbstractHttpControllerTestCase;
 
 class IndexControllerTest extends AbstractHttpControllerTestCase
 {
-    public function setUp()
+    public function setUp(): void
     {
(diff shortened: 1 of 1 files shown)


webimpress:
@KOVCHENKO 

The problem is due to the latest release of zend-test (3.3.0) with support for PHPUnit 8.

When using installer you are choosing only to install zend-test and the latest version of PHPUnit 8 is installed because you are running PHP 7.2+.

Unfortunately, skeleton still supports PHP 5.6 so your change is not gonna work.
All will be fine if you downgrade PHPUnit or update the file in your project as you've suggested here.

Not sure how to do it in the skeleton, so it works on all versions (PHP 5.6, PHP 7.X) with PHPUnit 4,5,6,7,8... I can see only tricky way, but I would like to avoid it in the skeleton.

/cc @Xerkus

weierophinney:
Unfortunately, since the skeleton still supports PHP 5.6, we cannot accept this change yet.

PR #452, opened 2019-11-27
Use script for updating .gitignore

### LINKED ISSUE #450: composer scripts is not compatible with windows
composer.json:
in "scripts": { "post-create-project-cmd": [...] }

please change double quoting to single quoting in the line:

"php -r \"$file = file_get_contents('.gitignore'); $file = str_replace('composer.lock', '', $file); file_put_contents('.gitignore', $file);\""

weierophinney:
Due to quoting issues between different operating systems, the script for removing the `composer.lock` entry from the `.gitignore` file was not working on Linux after a recent update to provide support for Windows. To provide a truly cross-platform way to work, this patch provides a script, `bin/update-gitignore`, which does the work. The `composer.json` entry is updated to call that script, which deletes itself on completion.

Fixes #450

### DIFF
--- a/bin/update-gitignore.php
+++ b/bin/update-gitignore.php
@@ -1,0 +1,12 @@
+<?php
+/**
+ * @see       https://github.com/zendframework/ZendSkeletonApplication for the canonical source repository
+ * @copyright Copyright (c) 2019 Zend Technologies USA Inc. (https://www.zend.com)
+ * @license   https://github.com/zendframework/ZendSkeletonApplication/blob/master/LICENSE.md New BSD License
+ */
+
+$gitIgnore = sprintf('%s/.gitignore', realpath(dirname(__DIR__)));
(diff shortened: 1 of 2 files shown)


bin/update-gitignore.php
```diff
@@ -0,0 +1,12 @@
+<?php
+/**
+ * @link      https://github.com/zendframework/ZendSkeletonApplication for the canonical source repository
+ * @copyright Copyright (c) 2019 Zend Technologies USA Inc. (https://www.zend.com)
+ * @license   https://framework.zend.com/license/new-bsd New BSD License
```
michalbundyra: we are using link to LICENSE.md of the repository - as the above does not work

weierophinney: Updated.

michalbundyra (approved):
🚢

